PR #20, opened 2014-08-08
Add path package

trobert2:
For testing purposes we need to get the long path of filesystem paths
on Windows machines. By default non-UTF8 syscalls will return the short path
(Ex:"C:\PROGRA~1" instead of "C:\Program Files"). This package exports
previously private functions used in the symlink package to determine the
long path of a file/directory.


mattyw:
Is this still needed?

anastasiamac:
There have not been a response in more than a year.

I am closing this proposal, please re-submit rebased version if needed.

PR #96, opened 2014-12-02
Chmod abstraction

perrito666:
Chmod presents a problem on windows for all its Go implementations:
When File.Chmod is used it will panic and os.Chmod will seem to
work but in reality corrupt the file permissions by trying to do the same
attribute writing that in *nixes.
This implementation uses icacls to set equivalent ACLs to the requested
permission. As discussed previously with @fwereade it covers the permissions
currently used in juju and it should be expanded for each new requirement.
Since there is no direct mapping from one permissions system to the other
a generic imlementation would be dangerous and could lead to unnoticed
wrong permissions settings.

(Review request: http://reviews.vapour.ws/r/567/)

### DIFF
--- a/fs/chmod.go
+++ b/fs/chmod.go
@@ -8,7 +8,7 @@
 // Use of Chmod form either os or File will provoke your code to
 // panic or misvehave in windows.
 
-// -build windows
+// +build !windows
 
 package fs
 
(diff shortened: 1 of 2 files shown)


mattyw:
@perrito666 looks like this pr can be closed?

anastasiamac:
Discarded on Review Board.

PR #106, opened 2015-01-22
Empty feature flag is always enabled.

howbazaar:
This makes other code using it easier.

(Review request: http://reviews.vapour.ws/r/787/)


axw:
LGTM

PR #110, opened 2015-02-05
Corrected licence AGPL3->LGPL3

perrito666:
Corrected wrong licencing details that where copied from other projects.
All involved parties where asked for permission.

(Review request: http://reviews.vapour.ws/r/873/)


perrito666:
$$fixes-1416436$$

PR #113, opened 2015-02-06
Corrected licence AGPL3->LGPL3

perrito666:
Several files had the wrong licence due to copy and paste from the wrong place, the licence headers where corrected.Several files had the wrong licence due to copy and paste from the wrong place, the licence headers where corrected.

(Review request: http://reviews.vapour.ws/r/887/)


dimitern:
:shipit: LGTM

PR #114, opened 2015-02-23
debugstatus: add Duration to result

rogpeppe:
This will allow us to get some idea of how long each status took.

### DIFF
--- a/debugstatus/status_test.go
+++ b/debugstatus/status_test.go
@@ -23,6 +23,7 @@ var _ = gc.Suite(&statusSuite{})
 
 func makeCheckerFunc(key, name, value string, passed bool) debugstatus.CheckerFunc {
 	return func() (string, debugstatus.CheckResult) {
+		time.Sleep(time.Microsecond)
 		return key, debugstatus.CheckResult{
 			Name:   name,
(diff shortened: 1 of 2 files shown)


mhilton:
:+1: LGTM

debugstatus/status_test.go
```diff
@@ -37,6 +38,13 @@ func (s *statusSuite) TestCheck(c *gc.C) {
 		makeCheckerFunc("check2", "check2 name", "value2", false),
 		makeCheckerFunc("check3", "check3 name", "value3", true),
 	)
+	for key, r := range results {
+		if r.Duration < time.Microsecond {
```
bac: If I were doing this test in Python I'd mock time.Now and time.Since to produce deterministic results. Putting a sleep in a test, even for a microsecond, has a certain code smell. Is there not a better way?

bac:
:+1: with a question about the test.

rogpeppe:
I considered mocking but since the logic is so simple, I thought that mocking that would detract more from the readability of the main code than it adds to our confidence that the logic is correct.

I'm willing to be swayed on this if anyone feels particularly strongly about it.

PR #116, opened 2015-03-13
Copy the functionality of path/filepath independent of the current OS.

ericsnowcurrently:
We use the functionality provided by the stdlib path/filepath package throughout juju to great effect.  The problem is that those functions are tied to the current operating system (runtime.GOOS).  There are cases where we want to use those same functions but targeting some other OS.  This patch facilitates such functionality via:

 * a set of helper functions for most of the path/filepath functions which take a path separator argument;
 * a new Renderer interface that exposes all the functions in path/filepath that don't relate to a concrete filesystem;
 * an implementation of Renderer for unix-like OSes and for Windows, both of which are almost entirely light wrappers around the separator-based helper functions;
 * a NewRenderer factory function that gives you the right Renderer implementation based on the OS name you give it;
 * all this is located in a package called "filepath".

The implementation of the helper functions (in stdlib*.go) is almost entirely copied from the Go stdlib source code.  The only difference is that hard-coded path separators have been changed to use the separator argument.  It's unfortunate that we have to copy-and-paste code like this, but there aren't any better alternatives.

(Review request: http://reviews.vapour.ws/r/1162/)

### DIFF
--- a/LICENSE.Go
+++ b/LICENSE.Go
@@ -1,0 +1,27 @@
+Copyright (c) 2010 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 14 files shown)


ericsnowcurrently:
$$merge$$

PR #118, opened 2015-03-20
Added package management package.

aznashwan:
Added a package that abstracts away differences between apt and yum and allows for easy extension for other package managers/distros.

The abstraction takes place on 3 major levels:
             - package management commands (PackageCommander interface).
             - running of the above package management commands (PackageManager interface).
             - packaging-related configuration options (PackagingConfigurer interface).

These packages mean to completely replace /utils/apt and provide all the necessary configuration information for any and all packaging-related operations undergone in Juju core.

NOTE: the only thing missing (from my point of view, at least) are the tests, which I'll get to writing after a couple of pre-emptive reviews.

(Review request: http://reviews.vapour.ws/r/1218/)


gabriel-samfira:
$$merge$$

PR #122, opened 2015-03-27
Fix the Go LICENCE file.

ericsnowcurrently:
This includes renaming the file so that it doesn't get compiled as a .go file on Windows (and other case-insensitive languages).

(Review request: http://reviews.vapour.ws/r/1301/)


ericsnowcurrently:
$$merge$$

PR #129, opened 2015-04-24
Added securestring package.

aznashwan:
A go implementation of Windows'SecureStrings.
A pure go implementation of this will be required for the Windows services module from core.

This module mimics the functionality of the ConvertFrom-SecureString PowerShell commandlet when used in the following context:

```
PS > $ss1 = ConvertTo-SecureString "Your string" -AsPlainText -Force
PS > $ss1
System.Security.SecureString
PS > ConvertFrom-SecureString $ss1
# The above yields the encrypted string in hexadecimal form which may be passed
# to this module's Decrypt() function in order to get the original string.

# Conversely, calling the Encrypt() function will yield an output
# which is perfectly compatible with ConvertTo-SecureString:
PS > "Output of Encrypt()" | ConvertTo-SecureString | $ss2
PS > $ss2
System.Security.SecureString
```

* Additional notes:
  1. SecureStrings in Windows, in it of themselves, are nothing more than plain strings stored in memory which is completely locked.
  2. ConvertFrom-SecureString uses the DAPI encryption technique to yield a hexadecimal-encoded string which encrypts the data from the locked memory.
  3. DAPI encrypts with the current user session in mind; therefore, decrypting ConvertFrom-SecureString's output can NOT be done by a different user than that which did the encryption in the first place, be it by the PowerShell commandlets or the methods in this module !!!
  4. Because of various user session parameters being used in the encryption process, both ConvertFrom-SecureString and this module's Encrypt() function yield varying output at each call, thus comparing the outputs of two different encryptions, albeit of the same input, will fail.

* Related documentation:
  * [Crypt32.dll::CryptProtectData](http://msdn.microsoft.com/en-us/library/windows/desktop/aa380261%28v=vs.85%29.aspx)
  * [Crypt32.dll::CryptUnprotectData](http://msdn.microsoft.com/en-us/library/windows/desktop/aa380882%28v=vs.85%29.aspx)
  * [Kernel32.dll::LocalFree](http://msdn.microsoft.com/en-us/library/windows/desktop/aa366730%28v=vs.85%29.aspx)
  * [System.Security.Cryptography.CAPI::CRYPTOAPI_BLOB](http://msdn.microsoft.com/en-us/library/windows/desktop/aa381414%28v=vs.85%29.aspx)

(http://reviews.vapour.ws/r/1486/)

### DIFF
--- a/securestring/securestring.go
+++ b/securestring/securestring.go
@@ -1,0 +1,161 @@
+// Copyright 2015 Canonical Ltd.
+// Copyright 2015 Cloudbase Solutions SRL.
+// Licensed under the AGPLv3, see LICENCE file for details.
+//
+// +build windows
+
+package securestring
+
+import (
+	"encoding/hex"
+	"strings"
+	"syscall"
+	"unsafe"
+
+	"github.com/juju/errors"
+)
+
+var (
+	cryptdll  = syscall.NewLazyDLL("Crypt32.dll")
+	kerneldll = syscall.NewLazyDLL("Kernel32.dll")
+
+	// Syntax:
+	// procProtectData.Call(inputBlob *blob, dataDescription *string,
+	//		optionalEntropy *blob, freeWorkingSpace *void, proptStruct
+	//		*CRYPTPROTECT_PROMPTSTRUCT, dwflags uint, outputBlob *blob)
+	//
+	// Parameters:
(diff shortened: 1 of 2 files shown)


mattyw:
@aznashwan It looks like this has been moved to a new repo. If so can you close this pr?

PR #130, opened 2015-04-30
Add NormCase and SplitSuffix to filepath.Renderer.

ericsnowcurrently:
NormCase is useful when dealing with Windows.  SplitSuffix is useful when dealing with file extensions.  Having these methods on Renderer helps with our cross-platform compatibility.  Both methods are inspired by Python's os.path module.

(Review request: http://reviews.vapour.ws/r/1537/)

### DIFF
--- a/filepath/common.go
+++ b/filepath/common.go
@@ -4,9 +4,9 @@
 package filepath
 
 func splitSuffix(path string) (string, string) {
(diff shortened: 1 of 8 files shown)


ericsnowcurrently:
$$merge$$

PR #132, opened 2015-05-07
Change shell.Renderer.RedirectOutput to append rather than truncate.

ericsnowcurrently:
Also add RedirectOutputReset for the truncate case.

(Review request: http://reviews.vapour.ws/r/1614/)


ericsnowcurrently:
$$merge$$

PR #133, opened 2015-05-11
Added support for detecting symlinks

bogdanteleaga:
This PR adds a IsSymlink function that is used in core to detect symlinks.

(Review request: http://reviews.vapour.ws/r/1643/)


bogdanteleaga:
$$merge$$

PR #134, opened 2015-05-26
exec: don't use stdin to send commands to shell

axw:
Using stdin to send commands to the shell means that commands run with this cannot use stdin, or they will consume subsequent script input. Instead, write a temporary script file and keep it around until the process is Waited upon.

Tested on Linux and Windows.

(Review request: http://reviews.vapour.ws/r/1781/)


axw:
$$merge$$

PR #135, opened 2015-05-26
featureflag: fix data race access flags

davecheney:
Fixes LP 1458717

(Review request: http://reviews.vapour.ws/r/1782/)


davecheney:
$$merge$$

PR #136, opened 2015-06-02
Add ChownPath

mjs:
This func sets the owner and group of a file to match the given username. This is needed in a couple of places in Juju so it is being extracted to here. It has not been implemented for Windows.

(Review request: http://reviews.vapour.ws/r/1838/)


mjs:
$$merge$$

PR #137, opened 2015-06-04
utils: make uuid parsing more lenient

rogpeppe:
This brings the UUID checking into line with the checking
done by UUID in juju/schema. There's no particular reason
to disallow other RFC4122 UUIDs. Once parsed
we treat the bytes as an opaque block, so whether it's
actually a version 4 UUID is immaterial.

(Review request: http://reviews.vapour.ws/r/1856/)

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -16,21 +16,17 @@ import (
 type UUID [16]byte
 
 // regex for validating that the UUID matches RFC 4122.
-// This package specifically generates and accepts version 4 UUIDs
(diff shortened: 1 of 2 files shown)


TheMue:
LGTM, even if it allows UUIDs not following the RfC. That's a different topic, like discussed on irc.

frankban:
:+1: thank you!

rogpeppe:
$$merge$$

PR #138, opened 2015-06-16
deque: allow a maximum length to be specified.

mjs:
Pushes which cause the length to exceed the specified size will cause an item to be dropped from the opposing side.

(Review request: http://reviews.vapour.ws/r/1940/)


mjs:
$$merge$$

PR #141, opened 2015-06-30
Trim whitespace out of keyvalues.

cmars:
Ran into a problem with one of the hook tools that would benefit from trimmed whitespace. This should address a whole class of potential problems with whitespace creeping into os.Args.

(Review request: http://reviews.vapour.ws/r/2064/)


cmars:
$$merge$$

PR #142, opened 2015-07-09
Fix licences

sinzui:
This branch fixes contradictory licence information that probably happened because files from moved from the juju/juju project, or people's editors are set to insert the wrong licence when creating a new file.

These fixes https://bugs.launchpad.net/juju-core/+bug/1472749 what was raised by a review of juju 1.24.2 for inclusion in Ubuntu. 1.24.2 can be included in Ubuntu if we can show that the upstream repo has fixed the contradictory information.

I also offer a Makefile that with rules to check the licence statements in go files. As this project was fixed for the same reasons a few months ago, I hope an easy check will prevent Ubuntu audits from find reasons to reject juju.

(Review request: http://reviews.vapour.ws/r/2135/)


sinzui:
$$merge$$

sinzui:
$$merge$$

PR #143, opened 2015-07-16
Add atomic MoveFile that does not overwrite.

cmars:
An atomic MoveFile that returns an error rather than overwrites the destination. This will be used to address [LP1475271](https://bugs.launchpad.net/juju-core/+bug/1475271).

(Review request: http://reviews.vapour.ws/r/2188/)


cmars:
$$merge$$

PR #144, opened 2015-07-17
cache: new package

rogpeppe:
This provides a simple caching mechanism. We're moving this
from the charmstore package internal packages so that
it can be used elsewhere too.

One change from the original: we allow arbitrary keys
rather than restricting to just strings.

(Review request: http://reviews.vapour.ws/r/2198/)

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -1,0 +1,165 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+// Package cache provides a simple caching mechanism
+// that limits the age of cache entries and tries to avoid large
+// repopulation events by staggering refresh times.
+package cache
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	"gopkg.in/errgo.v1"
+)
+
+// entry holds a cache entry. The expire field
+// holds the time after which the entry will be
+// considered invalid.
+type entry struct {
+	value  interface{}
+	expire time.Time
+}
+
+// Key represents a cache key. It must be a comparable type.
+type Key interface{}
+
+// Cache holds a time-limited set of values for arbitrary keys.
+type Cache struct {
+	maxAge time.Duration
+
+	// mu guards the fields below it.
+	mu sync.Mutex
+
+	// expire holds when the cache is due to expire.
+	expire time.Time
+
+	// We hold two maps so that can avoid scanning through all the
+	// items in the cache when the cache needs to be refreshed.
+	// Instead, we move items from old to new when they're accessed
+	// and throw away the old map at refresh time.
+	old, new map[Key]entry
+}
+
+// New returns a new Cache that will cache items for
+// at most maxAge.
+func New(maxAge time.Duration) *Cache {
+	// A maxAge is < 2ns then the expiry code will panic because the
+	// actual expiry time will be maxAge - a random value in the
+	// interval [0, maxAge/2). If maxAge is < 2ns then this requires
+	// a random interval in [0, 0) which causes a panic.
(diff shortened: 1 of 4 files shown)


cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+
+type entry struct {
+	value  interface{}
+	expire time.Time
+}
+
+// Key represents a cache key. It must be a comparable type.
+type Key interface{}
+
+// Cache holds a time-limited cache of values for arbitrary keys.
```
davecheney: time limited set ? of values ?

cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+
+	"gopkg.in/errgo.v1"
+)
+
+type entry struct {
```
davecheney: even though this is a private symbol, commenting the type and it's fields wound be nice.

cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+	expire time.Time
+	// We hold two maps so that can avoid scanning through all the
+	// items in the cache when the cache needs to be refreshed.
+	// Instead, we move items from old to new when they're accessed
+	// and throw away the old map at refresh time.
+	old, new map[Key]entry
+}
+
+// New returns a new Cache that will cache items for
+// at most maxAge.
```
davecheney: Please document the lower bounds of maxAge

davecheney:
LGTM with some documentation nits

cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+func New(maxAge time.Duration) *Cache {
+	// A maxAge is < 2ns then the expiry code will panic because the
+	// actual expiry time will be maxAge - a random value in the
+	// interval [0. maxAge/2). If maxAge is < 2ns then this requires
+	// a random interval in [0, 0) which causes a panic.
+	if maxAge < 2*time.Nanosecond {
```
mhilton: This is a bit clunky for a general purpose package I feel (I know I wrote it). I have thought previously about something like:
```go
// New returns a new cache that will cache items for at most maxAge.
// If f is non nil it will be called before adding a new item into the cache, 
// the item's expiry time will be derived from the return value.
func New(maxAge time.Duration, f func(time.Duration) time.Duration) {
...
}
```

what do you think?

cache/cache.go
```diff
@@ -0,0 +1,160 @@
...
+
+// Len returns the total number of cached entries.
+func (c *Cache) Len() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return len(c.old) + len(c.new)
+}
+
+// Evict removes the entry with the given key from the cache if present.
+func (c *Cache) Evict(key Key) {
```
mhilton: I'm not sure that this is sufficient. the entry might still be in old and not expired, at which point Evict doesn't evict anything.

mhilton:
LGTM, with some API thoughts.

cache/cache.go
rogpeppe: There are no lower bounds. It's rounded up to 2ns but that should affect the caller in practice.

cache/cache.go
rogpeppe: I think we can add something similar later if need be.

cache/cache.go
rogpeppe: excellent point. I've added a test, checked that it failed, then fixed the logic.

cache/cache.go
rogpeppe: done

cache/cache.go
rogpeppe: done

rogpeppe:
$$merge$$

PR #146, opened 2015-08-07
Expand an existing helper for dealing with OS env vars.

ericsnowcurrently:
This patch is based originally on mergeEnvironment in exec/exec.go.  It also draws from OsEnvSuite in the testing repo (osenv.go).  With these changes in place we will be able to use this code in OsEnvSuite.

(Review request: http://reviews.vapour.ws/r/2321/)


anastasiamac:
Discarded on Review Board.

PR #148, opened 2015-09-02
return new HTTP clients rather than aliasing http.DefaultClient

rogpeppe:
This makes the Get*HTTPClient functions return a new client
each time they are called, implementing the semantics described in the
doc comments.

This fixes bugs where callers were setting fields in the returned
clients and unintentionally setting http.DefaultClient attributes
as a result and hence breaking test isolation.

(Review request: http://reviews.vapour.ws/r/2569/)


rogpeppe:
$$merge$$

PR #149, opened 2015-09-10
bzr: new package

davecheney:
Move bzr utils package from juju/bzr to the utils repo.

A followup PR will remove the package from juju and adjust dependencies.tsv

(Review request: http://reviews.vapour.ws/r/2622/)


davecheney:
$$merge$$

PR #154, opened 2015-09-25
Added version to series translation.

anastasiamac:
We can only get version (e.g. 14.04) from series (e.g. trusty).

This proposal allows to get series from version.

(Review request: http://reviews.vapour.ws/r/2751/)


anastasiamac:
$$merge$$

PR #156, opened 2015-09-30
lp1501381 - don't panic with empty version string

cherylj:
Modified VersionSeries and SeriesVersion to just return
unknownSeries / Version error when passed an empty string.

(Review request: http://reviews.vapour.ws/r/2797/)


cherylj:
$$merge$$

PR #159, opened 2015-10-08
Added stale lock detection and removal logic.

dooferlad:
This fixes a bug where if Juju was restarted with a lock file in place (for example, the machine was rebooted) then the lock would remain but nothing held it. This logic will delete an existing lock file if the locks PID (now stored in the nonce as well as a UUID) doesn't point to the current process or the lock is older than the current process. This ensures that even if Juju starts with the same PID as it had previously we can still detect that the lock is stale.

(Review request: http://reviews.vapour.ws/r/2854/)


dooferlad:
$$merge$$

PR #163, opened 2015-10-14
Pass the clock dependency into the fslock.

howbazaar:
Make the dependency on a clock explicit and required when constructing a fslock.

(Review request: http://reviews.vapour.ws/r/2887/)


howbazaar:
$$merge$$

PR #165, opened 2015-10-23
debugstatus: add Handler

rogpeppe:
This code is already in two services and rather than add it to a third, we factor
it out to here.

(Review request: http://reviews.vapour.ws/r/2990/)

### DIFF
--- a/debugstatus/handler.go
+++ b/debugstatus/handler.go
@@ -1,0 +1,118 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package debugstatus
+
+import (
+	"net/http"
+
+	"gopkg.in/errgo.v1"
+
+	pprof "github.com/juju/httpprof"
+	"github.com/juju/httprequest"
+)
+
+// Version describes the current version of the code being run.
+type Version struct {
+	GitCommit string
+	Version   string
+}
+
+// Handler implements a type that can be used with httprequest.Handlers
+// to serve a standard set of /debug endpoints, including
+// the version of the system, its current health status
+// the runtime profiling information.
+type Handler struct {
+	// Check will be called to obtain the current health of the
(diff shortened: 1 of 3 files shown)


debugstatus/handler.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Handler implements a type that can be used with httprequest.Handlers
+// to serve a standard set of /debug endpoints, including
+// the version of the system, its current health status
+// the runtime profiling information.
+type Handler struct {
+	// Check will be called to obtain the current health of the
+	// system. It should return a map as returned from the
+	// Check function.
+	Check func() map[string]CheckResult
```
mhilton: I think the Handler shouldn't require you to set all of these things to use it. If Check of CheckPprofAllowed is nil then we should do something that isn't panicking. /debug/check could just return an empty set (not sure if that should be considered successful or not) CheckPprofAllowed could just deny access.

mhilton:
LGTM with one thought about default behaviour

debugstatus/handler.go
rogpeppe: done

debugstatus/handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+		Message: err.Error(),
+	}
+}
+
+type handlerSuite struct {
+}
+
+var _ = gc.Suite(&handlerSuite{})
+
+var errUnauthorized = errgo.New("you shall not pass!")
```
mattyw: cute - but "unauthorized" to make the error a bit more explicit

mattyw:
LGTM - wish I'd had this a few months ago

debugstatus/handler_test.go
rogpeppe: naa, i'm gonna go with cuteness. :)

rogpeppe:
$$merge$$ :shipit:

PR #166, opened 2015-10-26
fslock: add NewLockWithClock function; maintain NewLock as backwardly compatible

rogpeppe:
Because NewLock had been changed backwardly incompatibly, it broke
any clients that used it unnecessarily. This maintains the old API
(which is more convenient for most clients) but adds the new functionality
using a new entry point.

(Review request: http://reviews.vapour.ws/r/2996/)


rogpeppe:
$$merge$$

PR #169, opened 2015-10-27
Added a timer to the utils package

bogdanteleaga:
This PR adds a timer implementation to the utils package.

This is supposed to be used inside uniter in a PR that is coming soon™. This is mostly the reason for how the interface has been designed, since it has to send a signal on some channel after a while. It is mostly a big wrapper around golang's AfterFunc.

http://reviews.vapour.ws/r/3010/


bogdanteleaga:
$$merge$$

PR #174, opened 2015-11-14
Add an abstraction around os/exec.

ericsnowcurrently:
This patch is part of the work for fixing https://bugs.launchpad.net/juju-core/+bug/1382556.  It amounts to creating an abstraction around `os/exec` that we can use in juju.  The next steps are to update juju core to use this abstraction and then to switch to a new Exec implementation that runs the commands on a separate "forker" daemon.  Those will be addressed in separate patches.

Note that tests constitute about 1850 out of 2300 added lines, leaving roughly 550 lines of non-test code to review. :)

(Review request: http://reviews.vapour.ws/r/3141/)


anastasiamac:
Discarded on Review Board.

PR #175, opened 2015-11-17
series: add win2016 and nano series

bogdanteleaga:
This PR adds series support for win 2016 and nano.

One thing to be aware of (it also appears in a comment) is that every windows server version will most likely have a complementary nano version from now on, so there is some code to work along this.

(Review request: http://reviews.vapour.ws/r/3164/)


bogdanteleaga:
$$merge$$

PR #178, opened 2015-11-24
Fix race.

dooferlad:
Create channel at lock creation time.

(Review request: http://reviews.vapour.ws/r/3221/)


dooferlad:
$$merge$$

PR #179, opened 2015-11-25
Revert "Fix race."

davecheney:
Reverts juju/utils#178

7e742fc4821e2eb3f2b2d91ddcc1bcce3ae97021 was not correct. It fixed the data race but avoiding overwriting the field in createAliveFile but did not solve the underlying invariant failure caused by createAfterFile being called more than once.

(Review request: http://reviews.vapour.ws/r/3228/)


davecheney:
$$merge$$

PR #180, opened 2015-11-25
fslock alive file now has much better control

dooferlad:
* Turn on and off by channels
 * Retries when files can't be created or updated (normally because it
   is so early in the lock lifetime that the initial create/move process
   is still happening).
 * More comprehensive testing around new functionality

(Review request: http://reviews.vapour.ws/r/3238/)


dooferlad:
$$merge$$

PR #181, opened 2015-11-26
tar: ignore headers in UntarFiles

### LINKED ISSUE #160: UntarFiles extracts pax_global_header
When using UntarFiles on a tar created by git, it will extract pax_global_header as a file.
I'm not sure if this is a problem with the standard library or here.
See https://gcc.gnu.org/ml/gcc/2006-03/msg00052.html for more on that.

axw:
Don't process entries other than directories,
symlinks, and regular files.

Fixes #160

(Review request: http://reviews.vapour.ws/r/3251/)


axw:
$$merge$$

PR #183, opened 2015-12-11
series: add UpdateSeriesVersions

axw:
Add a function, UpdateSeriesVersions, which will
refresh the supported series by reading ubuntu.csv
on Ubuntu, and do nothing on other platforms. This
will be used when upgrading Juju: we will upgrade
the distro-info package, and then call this function
to read it in.

(Review request: http://reviews.vapour.ws/r/3373/)


axw:
$$merge$$

PR #184, opened 2015-12-16
utils/series: remove series.test.exe

axw:
Accidentally added test binary.

(Review request: http://reviews.vapour.ws/r/3397/)


axw:
Bogdan confirmed by email that this was added by accident and can be safely removed.

axw:
$$merge$$

PR #186, opened 2016-01-14
fixes bug #1515289

natefinch:
We werent setting the proxy field in the HTTP transport struct, which is what
the http Client uses to figure out how to proxy... so of course, we weren't
respecting the environment proxy.  Oddly enough, tools download is the only logic
using this code.  (However a later patch I'm going to land will cause the metric
sender to use this code as well, since it's rolling its own tls transport, and
making the same mistakes.

Added a TLS handshake timeout identical to the one used by the default http client,
so behavior will be consistent with the default.

For reference, this is the default transport used by the http client:

```go
var DefaultTransport RoundTripper = &Transport{
        Proxy: ProxyFromEnvironment,
        Dial: (&net.Dialer{
                Timeout:   30 * time.Second,
                KeepAlive: 30 * time.Second,
        }).Dial,
        TLSHandshakeTimeout: 10 * time.Second,
}
```
(from https://golang.org/pkg/net/http/#DefaultTransport)

(Review request: http://reviews.vapour.ws/r/3529/)

### DIFF
--- a/http.go
+++ b/http.go
@@ -11,7 +11,6 @@ import (
 	"net/http"
 	"strings"
 	"sync"
-	"time"
 )
 
 var insecureClient = (*http.Client)(nil)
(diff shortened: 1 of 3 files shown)


natefinch:
Note: I had to break this out into a go 1.2 fix and a go 1.3 fix, since go 1.2 does not support the tls timeout handshake field.  This way we'll automatically get the timeout when we build with go 1.3+, and it'll maintain the identical behavior with the default transport.

natefinch:
$$merge$$

PR #188, opened 2016-01-19
Track size and hash even when there is an error.

ericsnowcurrently:
Some readers/writers handle bytes in the same operation that they handle EOF.  This change accommodates them.

(Review request: http://reviews.vapour.ws/r/3572/)


ericsnowcurrently:
$$merge$$

PR #189, opened 2016-01-20
Simplify hashing and sizing.

ericsnowcurrently:
io.TeeReader() and io.MultiWriter() provide functionality which we had duplicated here.  This patch removes most of the duplication.  HashingWriter is left until juju/juju master stops using it (state/backups/create.go).

(Review request: http://reviews.vapour.ws/r/3579/)


ericsnowcurrently:
$$merge$$

PR #190, opened 2016-01-21
Fix LP 1536378

davecheney:
Fixes LP #1536378

The logic inside isAlive was wrong, it was never executing the wait loop
so if two independent processes called isAlive (via clean) then they would
both think the lock is stale and attempt to break it.

The number of loop iterations if the alive file cannot be read is set to
an arbitrary value which will limit the max delay in tying to lock a dead lock
to 300 seconds.

(Review request: http://reviews.vapour.ws/r/3585/)

### DIFF
--- a/fslock/fslock.go
+++ b/fslock/fslock.go
@@ -146,7 +146,7 @@ func (lock *Lock) isAlive(PID int) bool {
 	if PID == lock.PID {
 		return true
 	}
-	for misses := 0; misses < 0; misses++ {
+	for i := 0; i < 10; i++ {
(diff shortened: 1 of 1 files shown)


davecheney:
$$merge$$

davecheney:
Bullshit, https://bugs.launchpad.net/juju-core/+bug/1536477

$$merge$$

davecheney:
CI bot is fucked up, cannot tell time.

http://paste.ubuntu.com/14588528/

davecheney:
$$merge$$

PR #192, opened 2016-01-26
add config.RequiresBackports

tych0:
This will be used for the implementation of
https://github.com/juju/juju/pull/4191 the LXD container type.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/packaging/config/functions.go
+++ b/packaging/config/functions.go
@@ -29,10 +29,7 @@ func GetCloudArchiveSource(series string) (packaging.PackageSource, packaging.Pa
 }
 
 func RequiresBackports(series string, pkg string) bool {
-	backportPkgs, ok := backportsBySeries[series]
-	if !ok {
-		return false
-	}
+	backportPkgs := backportsBySeries[series]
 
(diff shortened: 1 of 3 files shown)


packaging/config/functions.go
```diff
@@ -27,3 +27,18 @@ func GetCloudArchiveSource(series string) (packaging.PackageSource, packaging.Pa
 		return configureCloudArchiveSourceUbuntu(series)
 	}
 }
+
+func RequiresBackports(series string, pkg string) bool {
+	backportPkgs, ok := backportsBySeries[series]
```
axw: This `ok` check is unnecessary, you'll just range over a nil slice below and then return false anyway.

axw:
LGTM with one suggestion.

tych0:
On Tue, Jan 26, 2016 at 03:46:17PM -0800, Andrew Wilkins wrote:
> LGTM with one suggestion.

Thanks, I've pushed a fix.

axw:
$$merge$$

PR #193, opened 2016-01-27
add MultiReaderSeeker

rogpeppe:
This enables the stitching together of sequential io.ReadSeekers
similar to the way that io.MultiReader works.

(Review request: http://reviews.vapour.ws/r/3646/)

### DIFF
--- a/multireaderseeker.go
+++ b/multireaderseeker.go
@@ -1,0 +1,118 @@
+package utils
+
+import (
+	"io"
+	"os"
+
+	"github.com/juju/errors"
+)
+
+type multiReaderSeeker struct {
+	readers []io.ReadSeeker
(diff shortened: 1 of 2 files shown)


jrwren:
:+1: seems useful.

multireaderseeker.go
```diff
@@ -0,0 +1,118 @@
...
+		r.sizes[r.index] = r.offset
+		r.index++
+		r.offset = 0
+		err = nil
+	}
+	return n, err
+}
+
+// Read implements io.Seeker.Seek. It can only be used to seek to the
+// start.
```
jrwren: It looks like it can be used to seek to location other than the start.

rogpeppe:
$$merge$$

PR #201, opened 2016-03-27
1533262 - add wily and xenial to supported series

cherylj:
For non ubuntu clients, wily and xenial need
to be added to the supported series list.

(Review request: http://reviews.vapour.ws/r/4347/)


cherylj:
$$merge$$

PR #203, opened 2016-03-31
exec: always return at least an empty ExecResponse

bogdanteleaga:
I've had some really fringe errors with some code that still tried to check the ExecResponse even though the err is there deep inside the uniter(and it might exist in other places). Point is it's safer to always return at least an empty ExecResponse.

(Review request: http://reviews.vapour.ws/r/4386/)


anastasiamac:
$$merge$$

PR #204, opened 2016-04-09
1568312 - fail install after first "not found" err

cherylj:
We should stop trying to install packages if the
error indicates that the package could not be
found / does not exist.

Otherwise, we will spend an extra 5 minutes during
bootstrap waiting on a package that will never
install.

This 5 extra minutes causes Juju's CI tests to
timeout and is a bad user experience.

(Review request: http://reviews.vapour.ws/r/4502/)

### DIFF
--- a/packaging/manager/apt.go
+++ b/packaging/manager/apt.go
@@ -11,6 +11,7 @@ import (
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


cherylj:
$$merge$$

cherylj:
$$merge$$

PR #212, opened 2016-05-04
Update per review, return quickly if cached

reedobrien:
Forward porting a fix made in the 1.25 branch.

(Review request: http://reviews.vapour.ws/r/4767/)


reedobrien:
$$merge$$

PR #214, opened 2016-05-09
ssh: Add Option.SetStrictHostKeyChecking()

mjs:
This change adds Support for all the openssh StrictHostKeyChecking values, instead of just "yes" or "no". In order to preserve compatibility with the pre-existing behaviour, the default value is "no".

This is needed for work being done to juju ssh/scp/debughooks where sometimes the option needs to be unset.

(Review request: http://reviews.vapour.ws/r/4794/)

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -18,6 +18,31 @@ import (
 	je "github.com/juju/errors"
 )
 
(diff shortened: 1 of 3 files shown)


mjs:
$$merge$$

PR #217, opened 2016-05-23
fslock: return underlying error from lock.Message

davecheney:
Previously an error retrieving the lock message would have been
interpreted as an empty lock message, and the error discarded. Rectify
this by returning the underlying error in addition to the lock message.

(Review request: http://reviews.vapour.ws/r/4881/)


davecheney:
$$merge$$

davecheney:
$$merge$$

PR #218, opened 2016-05-24
fslock: unexport lock.isLockHeld

davecheney:
Unexport isLockHeld, it is unused in juju and cannot be used safely as
any value reported by this function is stale.

(Review request: http://reviews.vapour.ws/r/4886/)


davecheney:
$$merge$$

PR #219, opened 2016-05-26
Use names.v2 and update dependencies

wallyworld:
We use gopkg.in/juju/names.v2 now.
The usage of names is in set.NewTagsFromStrings() which isn't called by core, so this won't have any transitive effects on core.

Also refresh to pull in more recent versions.

(Review request: http://reviews.vapour.ws/r/4905/)


wallyworld:
$$merge$$

PR #220, opened 2016-06-20
Extract natural sorting helpers & make more generic

mjs:
The sorting implementation which sorts strings with integer suffixes "naturally" has been extracted from github.com/juju/juju/cmd/juju/common. 

The implementation as been made generic so that it supports multiple runs of digits. In Juju's case this means that machine ids for containers are correctly sorted, as are unit names which contain digits before the unit number.

Testing of the algorithm has been simplified. Instead of having to provide an input and the desired output, only the desired output needs to be provided. The test infrastructure will test the algorithm by sorting a reverse of the input as well as a randomly shuffled version of the input. This makes writing test cases much easier.

Test cases have also been made more orthogonal. The unnecessarily long test names have also been cleaned up.

(Review request: http://reviews.vapour.ws/r/5108/)

### DIFF
--- a/naturalsort.go
+++ b/naturalsort.go
@@ -1,0 +1,57 @@
+// Copyright 2015 Canonical Ltd.
+// Licensed under the AGPLv3, see LICENCE file for details.
+
+package common
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"unicode"
+)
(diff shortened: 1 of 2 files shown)


mjs:
$$merge$$

mjs:
$$merge$$

PR #221, opened 2016-06-21
Remove fslock.

howbazaar:
Better replacement found at github.com/juju/mutex.

(Review request: http://reviews.vapour.ws/r/5129/)


howbazaar:
$$merge$$

PR #222, opened 2016-06-23
Add ConformYAML method from core.

howbazaar:
This method was in cmd/juju/common, but it is more useful than that.

(Review request: http://reviews.vapour.ws/r/5144/)


howbazaar:
$$merge$$

PR #226, opened 2016-07-26
fixes http://pad.lv/1604474 - userdata execution fails on Windows

natefinch:
Add support for two ciphersuites that Windows likes, since we accidentally
pared our supported ciphersuite list down to ones that Windows doesn't like
(and a few that aren't useful, since they're not RSA).

fixes http://pad.lv/1604474
(Review request: http://reviews.vapour.ws/r/5307/)


mitechie:
@jameinel can you please take a peek at this? 

@natefinch do we not need any sort of test that asserts that windows will in fact hit these ciphers and use them for something?

anastasiamac:
@mitechie 
re: "any sort of test" - we do have deployment tests in CI that also have also been failing according the bug log.

mitechie:
Right but the test failure is indirect correct? It seems reasonable if
we've got the cipherlist updated for Windows that there's a unit test that
would directly dail if the chipher was removed removing a bunch of
debugging work to find out why a change here demonstrates as a fail over
there.

natefinch:
I could maybe work out a unit test.  The problem is that it's powershell's list of usable ciphers that we care about, so I'd have to execute powershell from the unit test.  It's not the end of the world, but it's sort of annoying and possibly fragile.  I'll work on it, though, because it definitely is valuable.

natefinch:
Now with a test :)

howbazaar:
$$merge$$

PR #227, opened 2016-07-29
tar: Avoid error when file grows during copy

bz2:
Fixes lp:1457575 at the utils/tar level.

The symptom being seen in juju is "archive/tar: write too long"
during backup. This comes from the log file that is being added
to the tar being appended to between the point the size is taken
from stat and put in the tar header, and io.Copy finishing.

Simply using io.CopyN instead limiting the size written to what
the header specifies is sufficient to avoid the issue.

(Review request: http://reviews.vapour.ws/r/5334/)


bz2:
$$william_lgtm$$

PR #230, opened 2016-08-09
update tomb dependency

nskaggs:
In support of https://github.com/juju/juju/pull/5956

PR #231, opened 2016-08-10
update juju/cmd dependency to reflect the gofmt revision

nskaggs:
Per approval from #230 , this merely updates the dependency to select the gofmt'ed version of the change to juju/cmd.


nskaggs:
$$merge$$

PR #233, opened 2016-08-15
make outgoing address checker work under Go 1.7

rogpeppe:
In Go 1.7, http.Transport introduced the DialContext field
which overrides the Dial field and is set in http.DefaultContext.

This means that our hack to override the Dial field to prevent
outgoing network connections in test doesn't work from
Go 1.7 onwards.

This PR fixes that. We change the tests to avoid patching
the dial call because there is no longer a single entry point,
so we use more functional checks instead, and do some
more comprehensive testing of the local address checking
logic.

(Review request: http://reviews.vapour.ws/r/5438/)

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -10,11 +10,11 @@ import (
 var (
 	GOMAXPROCS        = &gomaxprocs
 	NumCPU            = &numCPU
-	Dial              = dial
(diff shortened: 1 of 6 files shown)


rogpeppe:
$$merge$$

PR #237, opened 2016-09-15
Replace Arch Linux support with "GenericLinux"

mjs:
All Linuxes which aren't Ubuntu or CentOS will be classified as GenericLinux This is part of the work to allowing the client to work on all Linux variants.

Part of the fix for: https://bugs.launchpad.net/juju/+bug/1616531

### DIFF
--- a/series/series_linux_test.go
+++ b/series/series_linux_test.go
@@ -168,17 +168,6 @@ VERSION_ID="7"
 	"centos7",
 	"",
 }, {
-	`NAME="Arch Linux"
-ID=arch
-PRETTY_NAME="Arch Linux"
-ANSI_COLOR="0;36"
-HOME_URL="https://www.archlinux.org/"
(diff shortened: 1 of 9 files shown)


series/series_linux_test.go
```diff
@@ -192,22 +181,46 @@ BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
...
-	"unknown",
+	"genericlinux",
 	"",
+}, {
+
+	"",
+	"unknown",
+	"OS release file is missing ID",
+}, {
+	`NAME="CentOS Linux"
```
howbazaar: Shouldn't we know about centos?

mjs: That test is checking what happens when the os-release file says "CentOS" but there's no information to determine the CentOS series (hence the "could not determine series" error)

mjs:
$$merge$$

PR #240, opened 2016-09-23
cache: single-flight fetch requests

rogpeppe:
This means that when there are many cache misses for the same key
at the same time, we don't get a thundering herd of fetches.


mhilton:
:+1:

fabricematrat:
:+1:

rogpeppe:
$$merge$$

PR #245, opened 2016-10-06
retry: new package

rogpeppe:
The plan is that this should eventually replace utils.Attempt
and github.com/juju/retry.

It provides functionality for stopping and clock mocking and for pluggable
strategies.

It is (arguably) quite a bit simpler to use than github.com/juju/retry and
the usage remains basically compatible with utils.Attempt, so
only minor code changes will be needed to change the existing
code base to use it.

### DIFF
--- a/retry/exp.go
+++ b/retry/exp.go
@@ -1,0 +1,50 @@
+package retry
+
+import (
+	"time"
+)
+
+// Exponential represents an exponential backoff retry strategy.
+// To limit the number of attempts or their overall duration, wrap
+// this in LimitCount or LimitDuration.
+type Exponential struct {
+	// Initial holds the initial delay.
+	Initial time.Duration
+	// Factor holds the factor that the delay time will be multiplied
+	// by on each iteration.
+	Factor float64
+	// MaxDelay holds the maximum delay between the start
+	// of attempts. If this is zero, there is no maximum delay.
+	MaxDelay time.Duration
+}
+
+type exponentialTimer struct {
+	strategy Exponential
+	start    time.Time
+	end      time.Time
+	delay    time.Duration
+}
+
+// NewTimer implements Strategy.NewTimer.
+func (r Exponential) NewTimer(now time.Time) Timer {
+	return &exponentialTimer{
+		strategy: r,
+		start:    now,
+		delay:    r.Initial,
+	}
+}
+
+// NextSleep implements Timer.NextSleep.
+func (a *exponentialTimer) NextSleep(now time.Time) (time.Duration, bool) {
+	sleep := a.delay - now.Sub(a.start)
+	if sleep <= 0 {
+		sleep = 0
+	}
+	// Set the start of the next try.
+	a.start = now.Add(sleep)
+	a.delay = time.Duration(float64(a.delay) * a.strategy.Factor)
+	if a.strategy.MaxDelay > 0 && a.delay > a.strategy.MaxDelay {
+		a.delay = a.strategy.MaxDelay
+	}
+	return sleep, true
+}
--- a/retry/regular.go
+++ b/retry/regular.go
@@ -1,0 +1,67 @@
+// Copyright 2016 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package retry
+
+import (
+	"time"
+
+	"github.com/juju/utils/clock"
+)
+
+// Regular represents a strategy that repeats at regular intervals.
+type Regular struct {
+	// Total specifies the total duration of the attempt.
+	Total time.Duration
+
+	// Delay specifies the interval between the start of each try
+	// in the burst. If an try takes longer than Delay, the
+	// next try will happen immediately.
+	Delay time.Duration
+
+	// Min holds the minimum number of retries. It overrides Total.
+	// To limit the maximum number of retries, use LimitCount.
+	Min int
+}
+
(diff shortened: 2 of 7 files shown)


retry/strategy.go
```diff
@@ -0,0 +1,59 @@
+package retry
+
+import (
+	"time"
+)
+
+type strategyFunc func(now time.Time) Timer
```
axw: I love the composable strategies.

retry/retry.go
```diff
@@ -0,0 +1,108 @@
...
+	waited  bool
+	running bool
+}
+
+// Start begins a new sequence of attempts for the given strategy. If
+// clk is nil, clock.WallClock will be used. If a value is received on
+// stop while waiting, the attempt will be aborted.
+func Start(strategy Strategy, clk clock.Clock, stop <-chan struct{}) *Attempt {
+	if clk == nil {
+		clk = clock.WallClock
```
rogpeppe: I'd like this package to be useful for one-off programs too and requiring every such use
to have an explicit dependency on juju/utils/clock seems a bit wrong to me.
If people are using a mock clock, they'll pass it in here. If they're not, I don't think
typing "clock.WallClock" is going to change matters much.

retry/retry.go
```diff
@@ -0,0 +1,142 @@
...
+type Timer interface {
+	// NextSleep returns the length of time to sleep
+	// before the next retry. If no more attempts should
+	// be made it should return false, and the returned
+	// duration will be ignored.
+	//
+	// Note that NextSleep is called once after
+	// each iteration has completed, assuming the
+	// retry loop is continuing.
+	NextSleep(now time.Time) (time.Duration, bool)
```
natefinch: When would you ever call this with anything other than time.Now()?  Isn't this why we have a fake clock implementation?

retry/exp.go
```diff
@@ -0,0 +1,57 @@
...
+
+type exponentialTimer struct {
+	strategy Exponential
+	start    time.Time
+	end      time.Time
+	delay    time.Duration
+}
+
+// Start is short for Start(r, clk, stop).
+func (r Exponential) Start(clk clock.Clock, stop <-chan struct{}) *Attempt {
```
natefinch: ew. If we can avoid the default case being (nil, nil), that would be really nice. Why not make the clock part of the struct? Surely no one is going to need to change out the clock when reusing an attempt strategy.

Then you could just have Start and StartWithCancel or something.

retry/retry.go
```diff
@@ -0,0 +1,142 @@
...
+		a.count++
+	case <-a.stop:
+		a.running = false
+		a.stopped = true
+	}
+	return a.running
+}
+
+// More reports whether the next call to Next will sleep again to wait
+// for another attempt. It does not itself block.
```
natefinch: This comment is really confusing. The firs thing it says is that it'll report if Next will sleep... .but that's an implementation detail.  What it reports is whether or not the attempt will retry.    I think that also indicates the name is not very good.  I don't know what More() means in the course of a an Attempt.  If it was called WillRetry, it would be very clear.

retry/retry.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"time"
+
+	"github.com/juju/utils/clock"
+)
+
+// Strategy is implemented by types that represent a retry strategy.
+type Strategy interface {
+	NewTimer(now time.Time) Timer
```
natefinch: This type needs a much bigger explanation of how someone would implement a Strategy.  I have no idea what NewTimer is supposed to return.

retry/regular.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+
+	"github.com/juju/utils/clock"
+)
+
+// Regular represents a strategy that repeats at regular intervals.
+type Regular struct {
+	Total time.Duration // total duration of attempt.
+	Delay time.Duration // interval between each try in the burst.
+	Min   int           // minimum number of retries; overrides Total
```
natefinch: I find it weird to have a *minimum* number of retries, and not a maximum.  I get it, in combination with the time limit, but still, I'd also expect there to be a way to say "just try 3 times"... I guess min works for that, if Total is zero.... but that's really counterintuitive... and doesn't give you the way to say "run for a max of Total time or max retries, whichever comes first"

Also, "overrides Total" is not exactly correct  Override, to me, means "totally ignore" ... but what actually happens is, you run for a min number of retries or for Total time, whichever comes later.

natefinch: > the Regular API is exactly the old AttemptStrategy API 

But this is a brand new package, so we should take the opportunity to make things less confusing.

> You can already do that with LimitCount. For example:

But that's a completely different mechanism than the minimum, which is why it's confusing.  It's not consistent.... so, for example, when someone looks at the fields of AttemptStrategy, they see minimum, but no maximum and think there's no way to do a maximum.  And if there's retry.LimitCount, why not retry.MinCount?

rogpeppe: > But that's a completely different mechanism than the minimum, which is why it's confusing.

Yes, it's a different mechanism because you can't do the min count thing by strategy composition (what timing would the extra events get?). We could use MinCount too, but I honestly don't think that name is confusing as it's in the context of the Regular struct type and documented.

I've updated the field documentation for Regular to use a more conventional comment style
and to mention LimitCount.

rogpeppe:
This package now lives at [gopkg.in/retry.v1](https://godoc.org/gopkg.in/retry.v1).

PR #247, opened 2016-10-31
Better support for GetOSFromSeries

wallyworld:
When calling GetOSFromSeries(), on Linux, if the series was not found, there was no call to update distro info data to pull in the last series info. This is fixed.

Also, add in yakkety and zesty so there work without the need for the distro info lookup.


wallyworld:
$$merge$$

PR #248, opened 2016-11-01
Allow MustHostSeries to be patched for testing

wallyworld:
For now, MustHostSeries needs to be a var so it can be patched for testing.

### DIFF
--- a/series/series.go
+++ b/series/series.go
@@ -21,6 +21,9 @@ const (
 
 var (
 	// TODO(katco): Remove globals (lp:1633571)
+	// Override for testing.
+	MustHostSeries = mustHostSeries
+
 	seriesOnce sync.Once
 	// These are filled in by the first call to hostSeries
 	series    string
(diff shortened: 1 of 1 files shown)


series/series.go
```diff
@@ -22,6 +22,7 @@ const (
 var (
 	// TODO(katco): Remove globals (lp:1633571)
 	seriesOnce sync.Once
+	MustHostSeries = mustHostSeries
```
anastasiamac: Please have it in a separate block from un-exported vars. Descriptive comment is missing too

series/series.go
```diff
@@ -41,7 +42,7 @@ func HostSeries() (string, error) {
 }
 
 // MustHostSeries calls HostSeries and panics if there is an error.
```
anastasiamac: Case needs to be changed here

series/series.go
wallyworld: it pertains to the todo

wallyworld:
$$merge$$

PR #249, opened 2016-11-01
This patch provides a friendly interface with the masterzen/winrm client

hoenirvili:
For a better integration in the juju ecosystem enabling many windows
machine interactions. This will also help us execute remote commands and scripts
across the wire.

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -1,4 +1,5 @@
 // Copyright 2012, 2013 Canonical Ltd.
+// Copyright 2016 Cloudbase solutions
 // Licensed under the AGPLv3, see LICENCE file for details.
 
 package cert
@@ -10,6 +11,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/pem"
 	"fmt"
 	"math/big"
@@ -19,88 +21,212 @@ import (
 	"github.com/juju/errors"
 )
 
-var KeyBits = 2048
-
-// ParseCert parses the given PEM-formatted X509 certificate.
-func ParseCert(certPEM string) (*x509.Certificate, error) {
-	certPEMData := []byte(certPEM)
-	for len(certPEMData) > 0 {
-		var certBlock *pem.Block
-		certBlock, certPEMData = pem.Decode(certPEMData)
-		if certBlock == nil {
-			break
-		}
-		if certBlock.Type == "CERTIFICATE" {
-			cert, err := x509.ParseCertificate(certBlock.Bytes)
-			return cert, err
-		}
-	}
-	return nil, errors.New("no certificates found")
+// OtherName type for asn1 encoding
+type OtherName struct {
+	A string `asn1:"utf8"`
 }
 
-// ParseCertAndKey parses the given PEM-formatted X509 certificate
-// and RSA private key.
-func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey, error) {
-	tlsCert, err := tls.X509KeyPair([]byte(certPEM), []byte(keyPEM))
-	if err != nil {
-		return nil, nil, err
-	}
-
-	cert, err := x509.ParseCertificate(tlsCert.Certificate[0])
-	if err != nil {
-		return nil, nil, err
-	}
-
-	key, ok := tlsCert.PrivateKey.(*rsa.PrivateKey)
-	if !ok {
-		return nil, nil, fmt.Errorf("private key with unexpected type %T", key)
-	}
-	return cert, key, nil
+// GeneralName type for asn1 encoding
+type GeneralName struct {
+	OID       asn1.ObjectIdentifier
+	OtherName `asn1:"tag:0"`
 }
 
-// NewCA generates a CA certificate/key pair suitable for signing server
-// keys for an environment with the given name.
-func NewCA(commonName, UUID string, expiry time.Time) (certPEM, keyPEM string, err error) {
-	key, err := rsa.GenerateKey(rand.Reader, KeyBits)
-	if err != nil {
-		return "", "", err
-	}
-	// TODO(perrito666) 2016-05-02 lp:1558657
-	now := time.Now()
+// GeneralNames type for asn1 encoding
+type GeneralNames struct {
+	GeneralName `asn1:"tag:0"`
+}
 
+// Config type used for specifing different params for NewLeaf func
+// This will effect the generation of certificates.
+type Config struct {
+	CommonName  string             // CommonName common name of the certificate
+	UUID        string             // UUID for a specific model
+	Expiry      time.Time          // Expiry when the certificate will expire
+	CA          []byte             // CA certifiacte authority to add a new leaf cert to it
+	CAKey       []byte             // CAKey private key of the CA to add a new leaf cert to it
+	IsCA        bool               // IsCA if we want to generate new a CA cert
+	Hostnames   []string           // Hostnames , list of hostnames for the certificate
+	ExtKeyUsage []x509.ExtKeyUsage // ExtKeyUsage extra flags for special usage of the cert
+	KeyBits     int                // KeyBits is used to set the lenght of the RSA key, default value 2048 bytes
+
+}
+
+// NewLeaf generates a certificate/key pair suitable for use
+// by a server, leaf node, client authentication, etc.
+// In order to generate certs for multiple purposes please consult the Config type.
+func NewLeaf(cfg *Config) (certPEM, keyPEM string, err error) {
+	var (
+		caCert *x509.Certificate
+		caKey  *rsa.PrivateKey
+	)
+
+	if cfg.CA != nil && cfg.CAKey != nil && !cfg.IsCA {
+		tlsCert, err := tls.X509KeyPair(cfg.CA, cfg.CAKey)
+		if err != nil {
+			return "", "", errors.Trace(err)
+		}
+		if len(tlsCert.Certificate) != 1 {
+			return "", "", fmt.Errorf("more than one certificate for CA")
+		}
+
+		caCert, err = x509.ParseCertificate(tlsCert.Certificate[0])
+		if err != nil {
+			return "", "", errors.Trace(err)
+		}
+		if !caCert.BasicConstraintsValid || !caCert.IsCA {
+			return "", "", errors.Errorf("CA certificate is not a valid CA")
+		}
+		var ok bool
+		caKey, ok = tlsCert.PrivateKey.(*rsa.PrivateKey)
+		if !ok {
+			return "", "", errors.Errorf("CA private key has unexpected type %T", tlsCert.PrivateKey)
+		}
+	}
+
+	// if none assign default
+	if cfg.KeyBits == 0 {
+		cfg.KeyBits = 2048
+	}
+
+	// generate private key
+	key, err := rsa.GenerateKey(rand.Reader, cfg.KeyBits)
+	if err != nil {
+		return "", "", errors.Errorf("cannot generate key: %v", err)
+	}
 	serialNumber, err := newSerialNumber()
 	if err != nil {
 		return "", "", errors.Trace(err)
 	}
-
-	template := &x509.Certificate{
-		SerialNumber: serialNumber,
-		Subject: pkix.Name{
-			CommonName:   commonName,
-			Organization: []string{"juju"},
-			SerialNumber: UUID,
-		},
-		NotBefore:             now.UTC().AddDate(0, 0, -7),
-		NotAfter:              expiry.UTC(),
-		SubjectKeyId:          bigIntHash(key.N),
-		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
-		IsCA:                  true,
-		MaxPathLen:            0, // Disallow delegation for now.
-		BasicConstraintsValid: true,
+	subject := pkix.Name{
+		CommonName:   cfg.CommonName,
+		Organization: []string{"juju"},
+		SerialNumber: cfg.UUID,
 	}
-	certDER, err := x509.CreateCertificate(rand.Reader, template, template, &key.PublicKey, key)
+	// get asn1 encoded info of the subject pkix
+	value, err := getUPNExtensionValue(subject)
 	if err != nil {
-		return "", "", fmt.Errorf("cannot create certificate: %v", err)
+		return "", "", fmt.Errorf("Can't marshal asn1 encoded %s", err)
 	}
+
+	// TODO(perrito666) 2016-05-02 lp:1558657
+	now := time.Now()
+	template := &x509.Certificate{
+		Subject:      subject,
+		SerialNumber: serialNumber,
+		NotBefore:    now.UTC().AddDate(0, 0, -7),
+		Version:      2,
+		NotAfter:     cfg.Expiry.UTC(),
+		SubjectKeyId: bigIntHash(key.N),
+		KeyUsage:     x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageKeyAgreement,
+		ExtKeyUsage:  cfg.ExtKeyUsage,
+		ExtraExtensions: []pkix.Extension{
+			{
+				Id:       subjAltName,
+				Critical: false,
+				Value:    value,
+			},
+		},
+	}
+
+	if cfg.IsCA {
+		template.BasicConstraintsValid = true
+		template.IsCA = true
+		template.KeyUsage = x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign
+	}
+
+	for _, hostname := range cfg.Hostnames {
+		if ip := net.ParseIP(hostname); ip != nil {
+			template.IPAddresses = append(template.IPAddresses, ip)
+		} else {
+			template.DNSNames = append(template.DNSNames, hostname)
+		}
+	}
+
+	if caKey == nil && caCert == nil {
+		caCert = template
+		caKey = key
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, template, caCert, getPublicKey(key), caKey)
+	if err != nil {
+		return "", "", err
+	}
+
 	certPEMData := pem.EncodeToMemory(&pem.Block{
 		Type:  "CERTIFICATE",
 		Bytes: certDER,
 	})
+
 	keyPEMData := pem.EncodeToMemory(&pem.Block{
 		Type:  "RSA PRIVATE KEY",
 		Bytes: x509.MarshalPKCS1PrivateKey(key),
 	})
 	return string(certPEMData), string(keyPEMData), nil
+}
+
+var (
+	// https://support.microsoft.com/en-us/kb/287547
+	//  szOID_NT_PRINCIPAL_NAME 1.3.6.1.4.1.311.20.2.3
+	szOID = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 311, 20, 2, 3}
+	// http://www.umich.edu/~x509/ssleay/asn1-oids.html
+	// 2 5 29 17  subjectAltName
+	subjAltName = asn1.ObjectIdentifier{2, 5, 29, 17}
+)
+
+// NewCA generates a CA certificate/key pair suitable for signing server
+// keys for an environment with the given name.
+func NewCA(commonName, UUID string, expiry time.Time, keyBits int) (certPEM, keyPEM string, err error) {
+	certPEM, keyPEM, err = NewLeaf(&Config{
+		CommonName: commonName,
+		UUID:       UUID,
+		Expiry:     expiry,
+		IsCA:       true,
+		KeyBits:    keyBits,
+	})
+	if err != nil {
+		return "", "", errors.Annotatef(err, "cannot generate ca certificate")
+	}
+	return
+}
+
+// NewClientCert generates a x509 client certificate used for https authentication sessions.
+func NewClientCert(commonName, UUID string, expiry time.Time, keyBits int) (certPEM string, keyPEM string, err error) {
+	certPEM, keyPEM, err = NewLeaf(&Config{
+		CommonName:  commonName,
+		UUID:        UUID,
+		Expiry:      expiry,
+		ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		KeyBits:     keyBits,
+	})
+	if err != nil {
+		return "", "", errors.Annotatef(err, "Cannot generate client certificate")
+	}
+
+	return
+}
+func bigIntHash(n *big.Int) []byte {
+	h := sha1.New()
+	h.Write(n.Bytes())
+	return h.Sum(nil)
+}
+
+// getUPNExtensionValue returns marsheled asn1 encoded info
+func getUPNExtensionValue(subject pkix.Name) ([]byte, error) {
+	// returns the ASN.1 encoding of val
+	// in addition to the struct tags recognized
+	// we used:
+	// utf8 => causes string to be marsheled as ASN.1, UTF8 strings
+	// tag:x => specifies the ASN.1 tag number; imples ASN.1 CONTEXT SPECIFIC
+	return asn1.Marshal(GeneralNames{
+		GeneralName: GeneralName{
+			// init our ASN.1 object identifier
+			OID: szOID,
+			OtherName: OtherName{
+				A: subject.CommonName,
+			},
+		},
+	})
 }
 
 // newSerialNumber returns a new random serial number suitable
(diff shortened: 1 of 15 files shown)


dooferlad:
Hi,

Thanks for the PR!

This is big and introduces new dependencies. Since at least https://github.com/masterzen/winrm is under a different license to juju/utils and juju/juju I would like to defer to someone who is able to work out what the implications are, if any, of incorporating them.

Please provide CI steps so the reviewer can confirm that the change works.

Please split the change up into smaller chunks of work. It seems that at least the changes in the series and shell folders could land independently of the winrm change.

The x509 certificate looks generic - is that right? If it is a specific winrm formatted certificate then I see the reason for including it in the winrm folder. Wouldn't combining this with github.com/juju/juju/cert/cert.go be useful? If so, moving code out of core into a security library is a good thing.

axw:
@hoenirvili I second @dooferlad's question about using juju/cert. I'm away this week, will review this when I get back.

gabriel-samfira:
@dooferlad The license for the winrm package is Apache 2.0 which should be consumable by juju/utils without any issues.

The x509 is mostly generic. The only winrm specific thing it does, is that it adds a UPN (User Principal Name) extension to the certificate. That extension is used to map the certificate to a particular system user (not unlike adding a OpenSSH key to the users ~/.ssh/authorized_keys).

Moving the x509 code to juju/cert should be fine. The reason we added it to the utils package was that we wanted to mirror the ssh package as closely as possible. Conceptually they have the same purpose (including the key generation part), so we wanted to have all the code dealing with access keys and remote commands in one place.

axw:
@gabriel-samfira another option is to split out the more generic parts of juju/cert, and move them into utils/cert (say). Then we can have winrm-specific bits inside winrm, and delegate to the generic bits in utils/cert.

gabriel-samfira:
@axw sounds good!

hoenirvili:
@axw  sounds good ! I will move the generic code from juju/cert to utils/cert.

hoenirvili:
@dooferlad the juju/cert is creating CA x509 certs but the winrm cert generation creates client certs to use for authentication purposes.

axw:
> @axw how does the manual provisioning (linux side) is integrated in the CI?

We have a subset of CI tests which run the exact same thing as for manual machines as for others. The only difference is that they use "add-machine ssh:..." and then deploy --to. We have some machines (I think LXD instances) running all the time, and each test run reprovisions the same machine.

For this branch, I don't think there are any QA steps required, as none of this is hooked up yet. Once winrm provisioning is hooked up, then QA steps will need to be listed describing how we can test your changes. e.g. start a windows VM, <prereq-steps-go-here>, juju add-machine winrm:...

hoenirvili:
@axw Could you take now another look?

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+// ErrAuth returned if the post request is droped due to invalid credentials
+var (
+	ErrAuth = fmt.Errorf("Unauthorized request")
+	ErrPing = fmt.Errorf("Ping failed, can't recive any response form target")
+)
+
+// Ping checks if we can make a WinRM conn use this client and runs a script
+// if the conn and script successfully runs it returns nil on call of Error()
+// If the default client of this pkg is not set this will panic.
+func (c *Client) Ping() {
```
axw: as with Run, I think this should just return the error directly

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+
+	logger.Debugf("Output of the ping maches %q", strings.Contains(stdout.String(), "HI!"))
+	if !strings.Contains(stdout.String(), "HI!") {
+		c.err = ErrPing
+		return
+	}
+}
+
+// UseClient set's the default client
+func UseClient(client Exec) {
```
axw: What value are the default client & top-level functions providing? I don't want to see them used in Juju, and this code is all about making winrm easier for Juju... so I'm inclined to think that we should drop the default client, UseClient, and other related top-level functions.

winrm/x509.go
```diff
@@ -0,0 +1,280 @@
...
+type x509 struct {
+	cert   []byte // client cert
+	key    []byte // client private key
+	cacert []byte // ca server cert
+	err    error  // internal err state
+}
+
+var (
+	mu          sync.Mutex
+	credentials x509 //winrm credentials
```
axw: I'm not keen on the use of globals here. I guess this is derived from the ssh/clientkeys.go code, which frankly is a bit ugly. The forces that directed ssh/clientkeys.go do not apply to winrm. In particular, we need keys for ssh *on disk* because we shell out to openssh. We don't do that for winrm.

Can you please change the code to parse and return certs/keys in-memory. I'd also prefer if we separated generation from loading. The caller can intercept an errors.IsNotFound(err), and generate a cert/key pair and write it to disk. I expect we'll only need to do that in one location in Juju, like we do for ssh.LoadClientKeys.

winrm/x509.go
```diff
@@ -0,0 +1,280 @@
...
+)
+
+// LoadClientCert generates client cert for x509 authentication
+// if the directory in the coresponding paths key, cert
+// if the base dir dosen't exist it will create and generate the key/cert pair
+// LoadClientCert function is lock protected allowing multiple threads to call this function
+func LoadClientCert(key, cert string) error {
```
axw: Please call them "keyFile", "certFile" or "keyPath", "certPath". Named as they are, I would expect them contain the key/cert values.

winrm/x509.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}()
+
+	return credentials.error()
+}
+
+// LoadCACert saves into internal singleton credentials the CACert to be used for https connections
+// if the cacert in the path is not found it will return the error rather than saing it into state
+// This behaviour will encourage that https connection will skip checking the CA.
+func LoadCACert(path string) error {
```
axw: As before, please don't use globals. It seems unlikely that we would have a well-defined location for a CA cert on disk, as it's going to be different for each server. In Juju, I guess there's two places where we would need to specify a CA cert:
 - when manual-provisioning, using a CA cert specified by the user on the command line
 - when doing the equivalent of "juju ssh" ("juju winrm"?), using a CA cert published by the Juju agent on the target machine

shell/powershell.go
```diff
@@ -51,3 +55,36 @@ func (pr *PowershellRenderer) MkdirAll(dirname string) []string {
...
+}
+
+// NewPSEncodedCommand converts the given string to a UTF16-LE, base64 encoded string,
+// suitable for execution using powershell.exe -EncodedCommand. This can be used on
+// local systems, as well as remote systems via WinRM.
+func NewPSEncodedCommand(script string) (string, error) {
+	var err error
+	script, err = newEncodedPSScript(script)
+	if err != nil {
+		return "", errors.Annotatef(err, "Can't construct powershell command for remote execution")
```
hoenirvili: Fixed !

shell/powershell_test.go
```diff
@@ -77,3 +77,17 @@ func (s powershellSuite) TestMkdirAll(c *gc.C) {
 		`mkdir 'C:\some\dir'`,
 	})
 }
+
+func (s powershellSuite) TestNewPSEncodedCommand(c *gc.C) {
```
hoenirvili: Fixed !

cert/cert.go
```diff
@@ -188,3 +180,163 @@ func bigIntHash(n *big.Int) []byte {
...
+			OID: szOID,
+			OtherName: OtherName{
+				A: subject.CommonName,
+			},
+		},
+	})
+}
+
+// NewClientCert generates a new x509 certificates based on the CertConfig passed
+func NewClientCert(commonName, UUID string, expiry time.Time) ([]byte, []byte, error) {
```
hoenirvili: Could you please add more details? Why should I add a pkix.Extensions to the New Leaf function? And why should I rewrite this function based on the New Leaf function?

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	Ping()
+	Run(string, io.Writer, io.Writer)
+	Password() string
+	Error() error
+	Secure() bool
+}
+
+// NewSecClient creates a new secure winrm client for initiating connections with the winrm listener
+// Note ,use this function before winrm.LoadClientKeys()
+func NewSecClient(user, host string) *Client {
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	cli := &Client{
+		secure: true,
+	}
+	// when creating a new client the winrm.DeafultParameters
+	// will be used to make a new client conneciton to a endpoint
+	// TransportDecorator will enable us to switch transports
+	// this will be used for https client x509 authentication
+	winrm.DefaultParameters.TransportDecorator = func() winrm.Transporter {
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	// - no CA if there is not any and yes if there is
+	// - pass only client cert and key
+	// - 30 sec timeout
+	if ca := CACert(); ca != nil {
+		endpoint = winrm.NewEndpoint(host, httpsPort, true, false, ca, ClientCert(), ClientKey(), 30*time.Second)
+	} else {
+		endpoint = winrm.NewEndpoint(host, httpsPort, true, true, nil, ClientCert(), ClientKey(), 30*time.Second)
+		logger.Warningf("Skipping CA server verification, using the --insecure option")
+	}
+	logger.Debugf("Creating WinRM secure clinet connection on %s %d", host, httpsPort)
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	// create winrm endpoint secure endpoint
+	// the endpoint has a target , host
+	// - secure port 5986
+	// - https connections are enabled
+	// - TLS accepts any certificate presented by the
+	// server and any host name in that certificate.
+	// - no CA if there is not any and yes if there is
+	// - pass only client cert and key
+	// - 30 sec timeout
+	if ca := CACert(); ca != nil {
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+
+	logger.Debugf("Creating WinRM clinet connection on %s port %d", host, httpPort)
+
+	// if we don't have defined user
+	// use the default administrator on windows machine
+	if user == "" {
+		user = defaultWinndowsUser
+	}
+
+	if gp != nil { // get password from a given way
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+		fmt.Fprintf(stdout, "[Winrm] >> ")
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	if cli.err != nil {
+		cli.err = errors.Annotatef(cli.err, "Connot create WinRM client conn")
+	}
+
+	return cli
+}
+
+// Run command and direct output to stdout and errors to stderr
+// If the Run successfully executes the cmd it returns nil on call of Error()
+func (c *Client) Run(command string, stdout io.Writer, stderr io.Writer) {
```
hoenirvili: If you look at the entire utils/winrm , the package is made with this principle in mind, of recording the error inside the client itself and future(client) operation return immediately if there was a previous error, making them no-op.  More on here: https://blog.golang.org/errors-are-values

winrm/winrm.go
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+// If the default client of this pkg is not set this will panic.
+func (c *Client) Ping() {
+	if c.err != nil {
+		return
+	}
+
+	logger.Debugf("Pinging WinRM listener")
+	var stdout, stderr bytes.Buffer
+
+	c.Run("ECHO HI!", &stdout, &stderr)
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	if stderr.Len() != 0 {
+		// save error
+		c.err = fmt.Errorf("%v", strings.TrimSpace(stderr.String()))
+		return
+	}
+
+	logger.Debugf("Output of the ping maches %q", strings.Contains(stdout.String(), "HI!"))
+	if !strings.Contains(stdout.String(), "HI!") {
```
hoenirvili: Fixed !

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+		c.err = ErrPing
```
hoenirvili: Fixed !

cert/cert.go
axw: AFAICS, the only difference we care about between NewClientCert and NewLeaf is that NewClientCert adds the UPN extension. So to avoid repeating all the generic certificate generation code here, I'd rather we used the NewLeaf to generate the cert, and provide the UPN extension as an argument to NewLeaf.

winrm/winrm.go
axw: I don't think the Scanner pattern is appropriate here. Run and Ping are not typically going to be used in a series, but as independent calls.

At the bottom of the blog article there is this comment, which is relevant:

> There is one significant drawback to this approach, at least for some applications: there is no way to know how much of the processing completed before the error occurred. If that information is important, a more fine-grained approach is necessary. Often, though, an all-or-nothing check at the end is sufficient.

winrm/winrm.go
```diff
@@ -0,0 +1,271 @@
...
+	if c.err != nil {
+		return
+	}
+
+	logger.Debugf("Pinging WinRM listener")
+	var stdout, stderr bytes.Buffer
+
+	c.Run("ECHO HI!", &stdout, &stderr)
+	if c.err != nil {
+		if strings.Contains(c.err.Error(), "401") {
```
hoenirvili: The errors that are coming back from winrm are not structured. That's why I check if 401(unauthorized request) is in the error string to know if we failed to access the endpoint because of unathorization problems or because the endpoint doesn't have winrm listener enable.

winrm/x509.go
```diff
@@ -0,0 +1,280 @@
...
+	"sync"
+	"time"
+
+	"github.com/juju/utils"
+	"github.com/juju/utils/cert"
+)
+
+// List of exported internal errors
+var (
+	ErrNoClientCert       = errors.New("No client winrm cert inside dir")
```
hoenirvili: Fixed !

winrm/x509.go
hoenirvili: Fixed !

winrm/winrm.go
hoenirvili: Ok @axw I see your reason, but please give me some advice, suggestions on how to make the API interactions better. We should also keep in mind that we need just only one client connection across the entire provisioning process. 
Something like:
  ```go
  conn := winrm.NewConn()
  doAllTheStuff(conn);
  winrm.Close(conn);
```

In my opinion I tried to do this using a global instance inside the utils/winrm package and make some wrapper methods around that, in order to make it more friendly I avoid using the api like``` winrm.ThatInstance.Method()```, instead we have. ```winrm.Method()``` and with methods like UseClient(), UseSecureClient() we are telling the client of the API to specify the method we should interact with the winrm listener, similarly we could introduces some global var to be set. Also if you saw, I was trying to mimic as much as I can the ```ssh pkg code```, to not make it radically different.

File: winrm/winrm.go (hunk 2 above)
Comment (hoenirvili) [inline winrm/winrm.go]: Check the upper last comment.

File: winrm/winrm.go (hunk 15 above)
Comment (axw) [inline winrm/winrm.go]: What I think would be natural is:

```go
package winrm

func NewClient(Config) *Client

type Client struct {...}

func (*Client) Close() error
func (*Client) Ping() error
func (*Client) Run(cmd string, stdout, stderr io.Writer) error
```

(I would prefer "Run" were "Command", like in ssh, but it may be overkill.)

Re mimicking the ssh package: I think the way the ssh package was done was a mistake, which I'd rather we didn't repeat. Maybe you can point me at some code that you think will be awkward if you get rid of the global?

hoenirvili: @axw if you look on the other PR(https://github.com/juju/juju/pull/6523) you can see I extensively use winrm.Run() after I initialise it in the ```InitAdministratorUser``` function, It would be a little unmaintainable to make NewClient() and Close() every time I run winrm.Run() into a function, I should not create more connections.We should just one connection for trying https auth vs password auth interactions for the entire provisioning process or other processes that will be used in the juju-core.

axw: @hoenirvili sorry, I was confused because there's a bit of interchanging of "Client" and "Connection" in this thread. Connection implies, to me, something that needs to be closed. The client isn't connection-oriented, though, so Close isn't necessary. So the API would just be:

```go
package winrm

func NewClient(Config) *Client

type Client struct {...}

func (*Client) Ping() error
func (*Client) Run(cmd string, stdout, stderr io.Writer) error
```

My previous statement about there being no sensible place for a single CA cert for all servers remains, so I still don't think it's sensible to have a global client. Not to mention it's racy: if you had two things in the same process trying to manually provision two different machines, we'd be in trouble. This isn't just academic, as we may want to move the bulk of manual provisioning responsibility to the controller sooner or later.

Seeing as you're passing in "DefaultClient" and "SecureClient" *anyway*, I see no reason why you would want to then pass that into a global and then use the global. Why not use the values directly? Doing so avoids concurrency issues, and makes testing easier.

Finally, on usage in PR juju/juju#6523. As above I was mistaken about the need for Close, so you needn't create a new Client per Run. Instead, just pass the pair of clients in (or a Client constructor function), and use them directly for the lifetime of the ProvisionMachine call. The Ping calls would be more idiomatic if the method returned an error, e.g.

at https://github.com/juju/juju/pull/6523/files#diff-8eee7a4e81e6a28a0cbc14fa50ae0b6bR128

```go
winrm.Ping()
first := winrm.Error()
if first == nil {
```

would become

```go
first := args.SecureClient.Ping()
if first == nil {
```

The Run calls would be slightly more verbose (i.e. with immediate error checking), but (a) there aren't that many Run calls, and (b) the errors will be directly relatable to the source (rather than "an error happened somewhere along the way", which you may or may not be able to determine the context of from the error message).

In case all of that is unclear, here is what I think InitAdministratorUser should look like:

```go
func InitAdministratorUser(args manual.ProvisionMachineArgs) error {
        logger.Infof("Initalising %q, user %q", args.Host, args.User)
        logger.Infof("Using WinRM secure client as user %q on %q", args.Host, args.User)
 
        first := args.SecureClient.Ping()
        if first == nil {
                logger.Infof("WinRM secure authentication is already enabled on the host %q", args.Host)
                return nil
        }
        logger.Debugf("Secure authentication is not enabled on user %q", args.User)
 
        if args.DefaultClient == nil { // if it's not already custom set use the default
                // at this state the secure listener is not enabled and
                // we must enable it first
                args.DefaultClient = winrm.NewClient(args.User, args.Host, winrm.TTYGetPasswd, args.Stdout)
        }

        last := args.DefaultClient.Ping()
        if last == nil {
                logger.Infof("WinRM normal connection is enabled on the host %q", args.Host)
                return nil
        }
        logger.Errorf("Normal authentication is not enabled on the host %q", args.Host)
 
        if first != nil && last != nil {
                return errors.Annotatef(first, last.Error())
        }
 
        if first != nil {
                return first
        }
 
        return last
}
```

(Bear in mind that ProvisionMachineArgs is a value type, so the update to DefaultClient won't persist beyond the end of the function. I think it might be better to require that ProvisionMachineArgs has a SecureClient and a DefaultClient on entry to the exported functions, using a Validate method.)

cert/cert.go
hoenirvili: Fixed !

winrm/winrm.go
hoenirvili: Fixed !

cert/cert.go
```diff
@@ -144,47 +181,145 @@ func NewLeaf(commonName, caCertPEM, caKeyPEM string, expiry time.Time, hostnames
...
 	}
-	certDER, err := x509.CreateCertificate(rand.Reader, template, caCert, &key.PublicKey, caKey)
+
+	if caCertPEM == "" &&
+		caKeyPEM == "" {
+		caCert = template
+		caKey = key
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, template, caCert, getPublicKey(key), caKey)
```
axw: Please update this function to call utils/cert.NewLeaf, instead of duplicating logic here. This can be in a follow up if you like.

cert/cert.go
```diff
@@ -144,47 +181,145 @@ func NewLeaf(commonName, caCertPEM, caKeyPEM string, expiry time.Time, hostnames
...
 	keyPEMData := pem.EncodeToMemory(&pem.Block{
 		Type:  "RSA PRIVATE KEY",
 		Bytes: x509.MarshalPKCS1PrivateKey(key),
 	})
 	return string(certPEMData), string(keyPEMData), nil
 }
 
+//NewClientCert generates a x509 client certificate used for https authentication sessions.
+func NewClientCert(commonName, UUID string, expiry time.Time) (string, string, error) {
+	cert, key, err := NewLeaf(commonName, "", "",
```
axw: Sorry, what I was asking for was for the function in this package to set up the ExtKeyUsage (any any other bits specific to winrm), and then call utils/cert.NewLeaf. There's a bunch of duplicated code in this package that isn't specific to winrm.

winrm/winrm.go
```diff
@@ -0,0 +1,235 @@
...
+
+// errNoPasswdFn sentinel for returning unset password callback
+var errNoPasswdFn = fmt.Errorf("No password callback")
+
+// password checks if the password callback is set and returns the password
+// from that specific get password handler
+func (c ClientConfig) password() (string, error) {
+	if c.Password != nil {
+		// make it look like a dropdown shell
+		fmt.Fprintf(os.Stdout, "[Winrm] >> ")
```
axw: please move the fmt.Fprintf prompts to the GetPasswd implementation. this code shouldn't assume GetPasswd is interactive.

winrm/winrm_test.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+func TestAll(t *testing.T) {
+	gc.TestingT(t)
+}
+
+type WinRMSuite struct{}
+
+var _ = gc.Suite(&WinRMSuite{})
+
```
axw: I think these tests all need to be updated now. There's no ResetClientCert, LoadClientCert, ClientCert, ClientKey, etc. Probably you should just have pre-generated certs/keys checked in the winrm_test package, and use them in your tests. Avoid generating new certs/keys for each test (or even the suite, if possible), because it's expensive and slows down tests. Not avoidable in all cases though.

cert/cert.go
hoenirvili: Fixed !

cert/cert.go
hoenirvili: Fixed !

hoenirvili:
Sorry for the delay @axw I will make a new commit.... somehow git doesn't' include the last fixes after rebasing with master,  I don't know what happened.

cert/cert.go
hoenirvili: Fixed !

winrm/winrm_test.go
hoenirvili: Fixed !

winrm/x509.go
```diff
@@ -0,0 +1,198 @@
+// Copyright 2016 Canonical Ltd.
+// Copyright 2016 Canonical Ltd.
```
axw: I think you want Cloudbase here.

winrm/x509.go
```diff
@@ -0,0 +1,198 @@
...
+	mu sync.Mutex
+
+	cert   []byte // client cert
+	key    []byte // client private key
+	cacert []byte // ca server cert
+
+}
+
+// NewX509 returns a new to an empty X509
+func NewX509() *X509 {
```
axw: The tests are all commented out, and this isn't used in the winrm package. Will you be using it in the juju code? If yes, please uncomment the tests and make sure they're working. If not, please delete the code.

cert/cert.go
axw: This is not fixed. Did you miss a commit? Maybe I'm being unclear.

What I want to see is:
 - no call to x509.CreateCertificate in this function
 - calling NewLeaf from github.com/juju/utils/cert (https://github.com/juju/utils/blob/master/cert/cert.go#L119) to create the certificate

To do that, github.com/juju/utils/cert would need to be updated so that you can pass x509.ExtraExtensions into NewLeaf. Let's just leave a TODO for now, and tidy that up later.

winrm/x509.go
hoenirvili: Fixed !

cert/cert.go
axw: Wait... this *is* the cert package. I think something was moved and I didn't notice, I'm sorry for the confusion.

axw: Hrm, I'm not so sure about changing the common utils/cert interface, but let's go with it for now. I may end up changing it a bit.

winrm/x509_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/juju/utils/winrm"
+
+	gc "gopkg.in/check.v1"
+)
+
+var (
+	base = "/tmp/base"
```
axw: please do not use a hard-coded temporary dir

winrm/x509_test.go
```diff
@@ -0,0 +1,61 @@
...
+	cert = "winrmcert.pem"
+	key  = "winrmkey.pem"
+)
+
+func (w *WinRMSuite) TestLoadClientCert(c *gc.C) {
+	cert := winrm.NewX509()
```
axw: `gocheck.C` has a method called MkDir that will create a temporary dir, and return its path. Please use that instead of the hard-coded "/tmp/base".

winrm/x509_test.go
hoenirvili: Fixed !

winrm/x509_test.go
hoenirvili: Fixed !

hoenirvili:
$$merge$$

hoenirvili:
$$merge$$

hoenirvili:
$$merge$$

PR #250, opened 2016-11-10
Move juju/cert pkg into utils.

hoenirvili:
As discussed in here https://github.com/juju/utils/pull/249
@axw @dooferlad

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -58,31 +58,9 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
 	return cert, key, nil
 }
 
-// Verify verifies that the given server certificate is valid with
-// respect to the given CA certificate at the given time.
-func Verify(srvCertPEM, caCertPEM string, when time.Time) error {
-	caCert, err := ParseCert(caCertPEM)
-	if err != nil {
-		return errors.Annotate(err, "cannot parse CA certificate")
-	}
-	srvCert, err := ParseCert(srvCertPEM)
-	if err != nil {
-		return errors.Annotate(err, "cannot parse server certificate")
-	}
-	pool := x509.NewCertPool()
-	pool.AddCert(caCert)
-	opts := x509.VerifyOptions{
-		DNSName:     "anyServer",
-		Roots:       pool,
-		CurrentTime: when,
-	}
-	_, err = srvCert.Verify(opts)
-	return err
-}
-
 // NewCA generates a CA certificate/key pair suitable for signing server
 // keys for an environment with the given name.
-func NewCA(envName, UUID string, expiry time.Time) (certPEM, keyPEM string, err error) {
+func NewCA(commonName, UUID string, expiry time.Time) (certPEM, keyPEM string, err error) {
 	key, err := rsa.GenerateKey(rand.Reader, KeyBits)
 	if err != nil {
 		return "", "", err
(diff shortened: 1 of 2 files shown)


cert/cert.go
```diff
@@ -0,0 +1,225 @@
...
+		Roots:       pool,
+		CurrentTime: when,
+	}
+	_, err = srvCert.Verify(opts)
+	return err
+}
+
+// NewCA generates a CA certificate/key pair suitable for signing server
+// keys for an environment with the given name.
+func NewCA(envName, UUID string, expiry time.Time) (certPEM, keyPEM string, err error) {
```
axw: Since this is meant to be generic, I think we should just pass in the common name and use it directly in the CommonName field. juju/cert can then have its own NewCA function which calls the one in this package with fmt.Sprintf(...)

cert/cert.go
```diff
@@ -0,0 +1,225 @@
...
+	key, ok := tlsCert.PrivateKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, nil, fmt.Errorf("private key with unexpected type %T", key)
+	}
+	return cert, key, nil
+}
+
+// Verify verifies that the given server certificate is valid with
+// respect to the given CA certificate at the given time.
+func Verify(srvCertPEM, caCertPEM string, when time.Time) error {
```
axw: unless this is needed by winrm, I think we should leave this one in juju/cert. It has some juju-specifics, because of the "anyServer".

cert/cert.go
```diff
@@ -0,0 +1,225 @@
...
+	return newLeaf(caCertPEM, caKeyPEM, expiry, hostnames, []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth})
+}
+
+// NewServer generates a certificate/key pair suitable for use by a server.
+func NewServer(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string) (certPEM, keyPEM string, err error) {
+	return newLeaf(caCertPEM, caKeyPEM, expiry, hostnames, []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth})
+}
+
+// newLeaf generates a certificate/key pair suitable for use by a leaf node.
+func newLeaf(caCertPEM, caKeyPEM string, expiry time.Time, hostnames []string, extKeyUsage []x509.ExtKeyUsage) (certPEM, keyPEM string, err error) {
```
axw: as above, I think this should take CommonName as an arg, and juju/cert can have its own version that passes in "*"

cert/cert.go
hoenirvili: This is not needed for winrm, winrm itself has built it function to verify the cert.

cert/cert.go
hoenirvili: Fixed !

cert/cert.go
hoenirvili: Fixed !

cert/cert.go
hoenirvili: Fixed !

cert/cert.go
```diff
@@ -0,0 +1,203 @@
...
+	// A serial number can be up to 20 octets in size.
+	// https://tools.ietf.org/html/rfc5280#section-4.1.2.2
+	n, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 8*20))
+	if err != nil {
+		return nil, errors.Annotatef(err, "failed to generate serial number")
+	}
+	return n, nil
+}
+
+// NewDefaultServer generates a certificate/key pair suitable for use by a server, with an
```
axw: can we just keep these two functions (NewDefaultServer and NewServer) in juju/cert, and expose newLeaf (NewLeaf)?

hoenirvili: Yeah sure ! Sounds reasonable.

hoenirvili:
$$merge$$

PR #251, opened 2016-11-17
ssh: Add StripCRReader for gocrypto on windows

bz2:
Fixes lp:1468752 by removing carriage returns from textual stdin
stream on windows when using gocrypto to communicate with remote
machine.

This is part one of making ssh from windows to a unix machine
more usable. Notably, this doesn't enable the passing of control
codes through to the remote shell, so for instance using a text
editor is still problematic.

QA steps
----

* Build juju using this version of utils for windows.
* Bootstrap with the resultant juju.exe
* `juju switch controller`
* `juju.exe ssh 0`
* On remote machine, `ls ~`
* See outout from ls and no other kipple.
* Run other commands as well, note `vi somefile` is still not usable.

### DIFF
--- a/ssh/ssh_gocrypto.go
+++ b/ssh/ssh_gocrypto.go
@@ -154,7 +154,7 @@ func (c *goCryptoCommand) ensureSession() (*ssh.Session, error) {
 	}
 	c.client = client
 	c.sess = sess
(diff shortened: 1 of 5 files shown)


bz2:
$$merge$$

PR #252, opened 2016-11-28
Decide when to use http or https port for winrm.

hoenirvili:
Forgot to add this lines of code in order to switch ports when we use HTTP port or HTTPS port.


hoenirvili:
$$build$$

PR #258, opened 2017-01-06
allow a cipher suite in common with standard gnutls

rogpeppe:
This should fix https://bugs.launchpad.net/juju-core/+bug/1654528
when incorporated into juju.


rogpeppe:
This is identical to PR 257, just targeted at 1.25.

rogpeppe:
$$merge$$

PR #261, opened 2017-01-18
Add function to return a public key from an x509 certificate

Pekkari:
The public key will be ready to be written in a config file. This is valuable to
generate a certificateduring the initial configuration of juju client that will
become a default certificate in case of trying to use lxd https remotes.

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -16,9 +16,11 @@ import (
 	"fmt"
 	"math/big"
 	"net"
+	"os"
 	"time"
 
 	"github.com/juju/errors"
+	"github.com/juju/utils"
 )
 
 // OtherName type for asn1 encoding
@@ -308,4 +310,19 @@ func ComputePublicKey(certPEM string) (string, error){
 		Bytes: marshalledPubKey,
 	})
 	return string(keyPEMData), nil
+}
+
+// StoreDefaultX509Cert takes a certificate and store it in the given
+// config location, using default names.
+func StoreDefaultX509Cert(location, pem, key, public string) (error) {
+	for file, content := range map[string]string{ "/juju-cert.pem": pem, "/juju-cert.key": key,
+						      "/juju-cert.pub": public } {
+		if _, err := os.Stat(location + file); os.IsNotExist(err) {
+			err := utils.AtomicWriteFile(location + file, []byte(content), 0600)
+			if err != nil {
+				return errors.Annotatef(err, "cannot write x509 certificate file: %s", location + file)
+			}
+		}
+	}
+	return nil
 }
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -292,3 +292,20 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
 	}
 	return cert, key, nil
 }
+
+// ComputePublicKey generates a public key from certPEM string
+func ComputePublicKey(certPEM string) (string, error){
+	cert, err := ParseCert(certPEM)
+	if err != nil {
+		return "", err
+	}
+	marshalledPubKey, err := x509.MarshalPKIXPublicKey(cert.PublicKey)
+	if err != nil {
+		return "", err
+	}
+	keyPEMData := pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PUBLIC KEY",
+		Bytes: marshalledPubKey,
+	})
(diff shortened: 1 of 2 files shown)


cert/cert.go
```diff
@@ -292,3 +292,18 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
 	}
 	return cert, key, nil
 }
+
+// Generate public key from keyPEM string
```
perrito666: the expected form for go doc is:
// ComputePublicKey generates public key from keyPEM string

cert/cert.go
```diff
@@ -292,3 +292,18 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
...
+func ComputePublicKey(certPEM string) (string, error){
+	cert, err := ParseCert(certPEM)
+	if err != nil {
+		return "", err
+	}
+	pubKey := cert.PublicKey.(interface {})
```
perrito666: sorry I dont fully understand the purpose of this typecasting.

cert/cert_test.go
```diff
@@ -219,6 +219,13 @@ func (certSuite) TestNewClientCertRSASize(c *gc.C) {
 	}
 }
 
+func (certSuite) TestComputePublicKey(c *gc.C) {
+	computedPublicKey, err := cert.ComputePublicKey(caCertPEM)
+	if err == nil {
```
perrito666: actually you want to c.Assert(err, gc.ErrorIsNil), if err is not nil this should fail too.

cert/cert.go
```diff
@@ -292,3 +292,18 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
...
+	marshalledPubKey, _ := x509.MarshalPKIXPublicKey(pubKey)
```
perrito666: what is the other ret value of MarshalPKIXPublicKey and why is it being ignored?

cert/cert.go
Pekkari: Sorry, I'm just a newbie in go, so I ignore this, I'll fix this.

cert/cert.go
Pekkari: That gets the public key as it's stored inside the certificate, and using the type that x509.MarshalPKIXPublicKey
expects in the input.

cert/cert.go
Pekkari: it is an error as usual, I didn't want to go so fine grained with several checks step after step, but if you need it, I certainly can add it.

cert/cert_test.go
Pekkari: Good one, I'll do, thanks!

cert/cert.go
perrito666: interface{} means that is a blank interface, absolutely anything will match this, you dont need a typecast, just pass whatever cert.PublicKey is and it will work.

Pekkari:
I'm heading to add another commit to this, as soon as I get some test case written, so don't be in a hurry to merge it.

cert/cert.go
```diff
@@ -292,3 +294,35 @@ func ParseCertAndKey(certPEM, keyPEM string) (*x509.Certificate, *rsa.PrivateKey
...
+	return string(keyPEMData), nil
+}
+
+// StoreDefaultX509Cert takes a certificate and store it in the given
+// config location, using default names.
+func StoreDefaultX509Cert(location, pem, key, public string) (error) {
+	for file, content := range map[string]string{ "/juju-cert.pem": pem, "/juju-cert.key": key,
+						      "/juju-cert.pub": public } {
+		if _, err := os.Stat(location + file); os.IsNotExist(err) {
+			err := ioutil.WriteFile(location + file, []byte(content), 0600)
```
axw: please use utils.AtomicWriteFile, otherwise the Stat/WriteFile pattern will be prone to half-written files

mjs:
@Pekkari: where are things at with this?

Pekkari:
Well, the juju pull request got stalled forever, if juju team doesn't like the idea of
generating a certificate on first run, there is no user for this code. Either we can wait
to them to finally reject the pull, or we can simple close it, as it's most likely not
going to proceed.

Thanks!

José.

axw:
Seeing as there's no pressing need for this any more, I'm going to close it. We can resurrect it if needed.

PR #262, opened 2017-01-25
Allow specifying a user to run script as.

howbazaar:
This is important for juju as the juju agents run as root, but we don't always want to execute commands as root. In particular when a user executes 'juju run' for a machine as opposed to an agent, we want to run those commands as the "ubuntu" user.

### DIFF
--- a/exec/exec.go
+++ b/exec/exec.go
@@ -113,7 +113,7 @@ func shellAndArgs(tempDir, script, user string) (string, []string, error) {
 				return "", nil, errors.Annotatef(err, "making tempdir readable by %q", user)
 			}
 			cmd = "/bin/su"
-			args = []string{user, "-c", fmt.Sprintf("/bin/bash %s", scriptFile)}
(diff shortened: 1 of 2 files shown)


howbazaar:
This has been personally QA'ed using a juju branch, but nothing particular here apart from running the tests.

exec/exec.go
```diff
@@ -101,10 +103,20 @@ func shellAndArgs(tempDir, script string) (string, []string, error) {
...
+			cmd = "/bin/bash"
+			args = []string{scriptFile}
+		} else {
+			// Need to make the tempDir readable by all so the user can see it.
+			err := os.Chmod(tempDir, 0755)
+			if err != nil {
+				return "", nil, errors.Annotatef(err, "making tempdir readable by %q", user)
+			}
+			cmd = "/bin/su"
+			args = []string{user, "-c", fmt.Sprintf("/bin/bash %s", scriptFile)}
```
mjs: Consider using --login

howbazaar:
$$merge$$

PR #263, opened 2017-02-03
mgokv: new package

rogpeppe:
This provides support for a global MongoDB-backed persistent
key-value store. It's optimised for reading - values are cached
for some period of time after being read or put.

### DIFF
--- a/mgopersist/export_test.go
+++ b/mgopersist/export_test.go
@@ -1,0 +1,10 @@
+// Copyright 2017 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package mgopersist
+
+var (
+	PutInitialAtTime = (*Session).putInitialAtTime
+	PutAtTime        = (*Session).putAtTime
+	GetAtTime        = (*Session).getAtTime
+)
--- a/mgopersist/persist.go
+++ b/mgopersist/persist.go
@@ -1,0 +1,190 @@
+// Copyright 2017 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+// Package mgopersist defines cached MongoDB-backed global persistent storage for
+// key-value pairs.
+package mgopersist
+
+import (
+	"sync"
+	"time"
+
+	"gopkg.in/errgo.v1"
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var ErrNotFound = errgo.New("persistent data entry not found")
+
+type entry struct {
+	// value holds the BSON-marshaled value. It will be empty
+	// if the entry was not found when fetched.
+	value []byte
+	// expire holds when the value will expire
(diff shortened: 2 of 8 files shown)


mgopersist/export_test.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2017 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package mgopersist
```
rogpeppe: done

mgopersist/persist.go
```diff
@@ -0,0 +1,190 @@
...
+// Package mgopersist defines cached MongoDB-backed global persistent storage for
+// key-value pairs.
+package mgopersist
+
+import (
+	"sync"
+	"time"
+
+	"gopkg.in/errgo.v1"
+	mgo "gopkg.in/mgo.v2"
```
rogpeppe: goimports does that by default, and who am I to argue?

mgopersist/persist.go
```diff
@@ -0,0 +1,190 @@
...
+	"gopkg.in/mgo.v2/bson"
+)
+
+var ErrNotFound = errgo.New("persistent data entry not found")
+
+type entry struct {
```
rogpeppe: I started changing it to use that package, but I'm not convinced it's entirely
appropriate here. Most of the complexity of utils/cache comes from dealing
with a potentially unbounded set of keys, which is something I don't think
we need to deal with here. I think I prefer it being simpler and free of that
dependency for the time being.

mgopersist/persist.go
```diff
@@ -0,0 +1,190 @@
...
+	Key string `bson:"_id"`
+	// Value holds the value. We store it as a raw value
+	// so that it looks nice when looking at the collection directly.
+	Value bson.Raw `bson:"value"`
+}
+
+// Session returns a store session that uses the given
+// collection for storage. Each store entry is stored
+// in a document in the collection.
+func (s *Store) Session(coll *mgo.Collection) *Session {
```
rogpeppe: I changed the interface to pass the session instead of the collection in Session.

I'd prefer not to call Copy in Session, as we've mgosession to bound the number
of sessions and I think that, even though it's inconvenient, we should stick to
the convention of passing in the session that's created for the request.

mgopersist/persist.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return true, nil
+}
+
+// Get gets the value associated with the given key into the
+// value pointed to by valpt, which should be a pointer to
+// the same struct type used to put the value originally.
+//
+// If the value is not found, it returns ErrNotFound.
+func (s *Session) Get(key string, valpt interface{}) error {
```
rogpeppe: done

mgopersist/export_test.go
rogpeppe: done

rogpeppe:
$$merge$$

PR #264, opened 2017-02-08
Add monotonic Now()

wallyworld:
Provide a Now() function which provides a time in nanoseconds from a monotonic clock.

### DIFF
--- a/clock/monotonic/monotonic.go
+++ b/clock/monotonic/monotonic.go
@@ -1,0 +1,22 @@
+// Copyright 2017 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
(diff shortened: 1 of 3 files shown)


clock/monotonic/monotonic.go
```diff
@@ -0,0 +1,22 @@
...
+
+package monotonic
+
+import (
+	"time"
+	_ "unsafe"
+)
+
+//go:noescape
+//go:linkname nanotime runtime.nanotime
```
axw: I'm very wary of depending on runtime internals like this. We only *need* it on Linux, so why not just use the syscall?

wallyworld:
$$merge$$

PR #265, opened 2017-02-10
utils: drop gabriel-samfira/sys fork dependency

axw:
We're no longer using Go 1.2, so we can drop the fork.
Part of fixing https://bugs.launchpad.net/juju/+bug/1470820.


axw:
$$merge$$

PR #266, opened 2017-02-13
mgokv: add Update method

rogpeppe:
This allows atomic updates to a value.

### DIFF
--- a/mgokv/persist.go
+++ b/mgokv/persist.go
@@ -121,6 +121,34 @@ func (s *Session) PutInitial(key string, val interface{}) (bool, error) {
 	return s.putInitialAtTime(key, val, time.Now())
 }
 
+// Update updates the value using the MongoDB update operation
+// specified in update. The value is stored in the "value" field
+// in the document.
+//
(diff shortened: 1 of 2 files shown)


mgokv/persist.go
```diff
@@ -121,6 +121,31 @@ func (s *Session) PutInitial(key string, val interface{}) (bool, error) {
...
+//
+// will atomically increment the N value.
+//
+// This function returns the error with the same cause
+// that mgo.Collection.Update returns.
+func (s *Session) Update(key string, update interface{}) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if err := s.coll.UpdateId(key, update); err != nil {
+		return errgo.Mask(err, errgo.Any)
```
mhilton: should this not return ErrNotFound as opposed to mgo.ErrNotFound?

mhilton: you'll also want to test this case

rogpeppe: Good idea. Done.

rogpeppe:
$$merge$$

PR #267, opened 2017-02-14
Adjust precedence used for juju apt proxy conf

wallyworld:
Adjust the precedence used for the apt proxy conf file so that Juju's settings override any MAAS settings.


wallyworld:
$$merge$$

PR #268, opened 2017-03-17
Fix gofmt and govet

wupeka:
Fixed some issues with formatting/vetting, minor changes mostly.

### DIFF
--- a/packaging/commands/interface.go
+++ b/packaging/commands/interface.go
@@ -99,7 +99,6 @@ func NewPackageCommander(series string) (PackageCommander, error) {
 	default:
 		return NewAptPackageCommander(), nil
 	}
-	return nil, nil
 }
 
 // NewAptPackageCommander returns a PackageCommander for apt-based systems.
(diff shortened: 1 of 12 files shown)


packaging/commands/interface.go
```diff
@@ -99,7 +99,6 @@ func NewPackageCommander(series string) (PackageCommander, error) {
 	default:
 		return NewAptPackageCommander(), nil
 	}
-	return nil, nil
```
perrito666: I am pretty sure this return is needed, I am not sure all versions of the compiler are that smart, what version did you try this with?

packaging/config/interface.go
```diff
@@ -49,8 +49,6 @@ func NewPackagingConfigurer(series string) (PackagingConfigurer, error) {
 	default:
 		return NewAptPackagingConfigurer(series), nil
 	}
-
-	return nil, nil
```
perrito666: same as with previous return

packaging/commands/interface.go
wupeka: 1.6.3

wupeka: I've verified and even 1.2.1 thinks it's unreachable.

wupeka:
$$merge$$

PR #269, opened 2017-03-17
Add NoProxy settings to packaging/commands/apt

wupeka:
Things needed to support for apt-no-proxy setting in juju.


wupeka:
$$merge$$

PR #270, opened 2017-03-20
Add go fmt and go vet to make check

wupeka:
`make check` was missing go fmt and go vet runs, added (with go vet syntax the same as for juju/juju).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,10 @@
 PROJECT := github.com/juju/utils
+GOFMT := $(shell gofmt -l .| sed -e "s/^/ /g")
+
+.PHONY: check-licence check-go check
+
+check: check-licence check-go
+	go test $(PROJECT)/...
 
 check-licence:
 	@(fgrep -rl "Licensed under the LGPLv3" .;\
@@ -7,10 +13,8 @@ check-licence:
 		xargs -I {} echo FAIL: licence missed: {}
 
 check-go:
-ifneq ($(strip $(shell gofmt -l .)),)
-	$(error go fmt is sad)
+ifneq ($(strip $(GOFMT)),)
+	$(error go fmt is sad: $(GOFMT))
 endif
 	@(go tool vet -all -composites=false -copylocks=false .)
 
-check: check-licence check-go
-	go test $(PROJECT)/...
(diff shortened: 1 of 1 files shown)


wupeka:
!!test!!

wupeka:
!!testplease!!

Makefile
```diff
@@ -6,5 +6,11 @@ check-licence:
 		find . -name "*.go") | sed -e 's,\./,,' | sort | uniq -u | \
 		xargs -I {} echo FAIL: licence missed: {}
 
-check: check-licence
+check-go:
+ifneq ($(strip $(shell gofmt -l .)),)
+	$(error go fmt is sad)
```
jameinel: you might want to just follow the "verify.bash" model. Its actually really nice to be told what files gofmt is unhappy about.
Also, I don't think 'gofmt -l .' recurses into subdirectories.

I also wonder if we should be adding
.PHONY: check-go check-licence

(shouldn't it be spelled license ?)

wupeka: 1. Done
2. Fortunately gofmt -l . does, go fmt doesn't
3. I've added PHONY for all phony targets, also moved check to the top (so that it's default)

(depends on which side of the pacific are you :)

Makefile
```diff
@@ -1,10 +1,20 @@
 PROJECT := github.com/juju/utils
+GOFMT := $(shell gofmt -l .| sed -e "s/^/ /g")
```
jameinel: It seems a little odd to run this unconditionally. Is it possible to put this under check-go?

Makefile
```diff
@@ -1,10 +1,20 @@
...
 	@(fgrep -rl "Licensed under the LGPLv3" .;\
 		fgrep -rl "MACHINE GENERATED BY THE COMMAND ABOVE; DO NOT EDIT" .;\
 		find . -name "*.go") | sed -e 's,\./,,' | sort | uniq -u | \
 		xargs -I {} echo FAIL: licence missed: {}
 
-check: check-licence
-	go test $(PROJECT)/...
\ No newline at end of file
+check-go:
+	$(eval GOFMT := $(shell gofmt -l .| sed -e "s/^/ /g"))
```
jameinel: You moved it inside, but you forgot to delete the earlier line.

jameinel:
$$merge$$

PR #273, opened 2017-04-20
add Setenv function

rogpeppe:
This makes it easier to override environment variables
since just appending environment variables doesn't
work any more (and never did for non-Go programs).
See https://github.com/golang/go/issues/19877 for
background.

### DIFF
--- a/setenv.go
+++ b/setenv.go
@@ -1,0 +1,30 @@
+// Copyright 2013 Canonical Ltd.
+// Licensed under the LGPLv3, see LICENCE file for details.
+
+package utils
+import (
+	"strings"
+)
+
+// Setenv sets an environment variable entry in the given env slice (as
+// returned by os.Environ or passed in exec.Cmd.Environ) to the given
+// value. The entry must be in the form "x=y" where x is the name of the
+// environment variable and y is its value.
+//
+// If a value isn't already present in the slice, the entry is appended.
+//
+// The new environ slice is returned.
+func Setenv(env []string, entry string) []string {
+	i := strings.Index(entry, "=")
+	if i == -1 {
+		panic("no = in environment entry")
+	}
+	prefix := entry[0 : i+1]
+	for i, e := range env {
+		if strings.HasPrefix(e, prefix) {
+			env[i] = entry
+			return env
+		}
+	}
+	return append(env, entry)
+}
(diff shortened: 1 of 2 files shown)


setenv_test.go
```diff
@@ -0,0 +1,34 @@
...
+var _ = gc.Suite(&SetenvSuite{})
+
+var setenvTests = []struct {
+	set    string
+	expect []string
+}{
+	{"foo=1", []string{"foo=1", "arble="}},
+	{"foo=", []string{"foo=", "arble="}},
+	{"arble=23", []string{"foo=bar", "arble=23"}},
+	{"zaphod=42", []string{"foo=bar", "arble=", "zaphod=42"}},
```
mattyw: test for wrong input format

setenv.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// If a value isn't already present in the slice, the entry is appended.
+//
+// The new environ slice is returned.
+func Setenv(env []string, entry string) []string {
+	i := strings.Index(entry, "=")
+	if i == -1 {
+		panic("no = in environment entry")
+	}
+	prefix := entry[0 : i+1]
```
jameinel: formatting looks odd. Is that really how 'go fmt' formats it?

setenv_test.go
```diff
@@ -0,0 +1,34 @@
...
+	gc "gopkg.in/check.v1"
+
+	"github.com/juju/utils"
+)
+
+type SetenvSuite struct{}
+
+var _ = gc.Suite(&SetenvSuite{})
+
+var setenvTests = []struct {
```
jameinel: It seems odd to have these as a slice outside of the function given that they depend so concretely on the initial setup.

setenv_test.go
rogpeppe: done

rogpeppe:
$$merge$$

setenv.go
```diff
@@ -0,0 +1,30 @@
...
+// value. The entry must be in the form "x=y" where x is the name of the
+// environment variable and y is its value.
+//
+// If a value isn't already present in the slice, the entry is appended.
+//
+// The new environ slice is returned.
+func Setenv(env []string, entry string) []string {
+	i := strings.Index(entry, "=")
+	if i == -1 {
+		panic("no = in environment entry")
```
jameinel: Anything that would be actively used in production, should *really* not have a panic() in it, since that tends to have the property of crashing all of Juju instead of just having one worker get restarted.

setenv.go
rogpeppe: yeah, it is.

PR #271, opened 2017-04-12
Add function to export proxy settings as systemd config

wupeka:
ProxySettings.AsSystemdDefaultEnv exports proxy settings in a form understandable by systemd.

PR #275, opened 2017-04-21
proxy: do not add extra newline

rogpeppe:
PR #271 changed AsScriptEnvironment so that
it adds a trailing newline to the script lines.

Some juju tests failed as a result of this change, so revert
to the previous behaviour - the final newline isn't
needed.


rogpeppe:
$$merge$$

PR #276, opened 2017-04-24
ssh: host-key checking for crypto/ssh client

axw:
We introduce a host-key checker that mimics
the behaviour of the OpenSSH client. The
default "StrictHostKeyChecking" behaviour
is now for the crypto/ssh client to ask
(if a terminal is available, else fail),
and to leave off the option for openssh
to defer the decision to the openssh client
code.

We also make the host key algorithms
configurable. This will be used in a Juju
PR to limit which host key algorithms are
acceptable.

Note: this PR requires the builders to be
updated to Go 1.8 before it can land, as
the newer golang.org/x/crypto revision
uses the standard library context package.

### DIFF
--- a/ssh/ssh_gocrypto.go
+++ b/ssh/ssh_gocrypto.go
@@ -4,6 +4,8 @@
 package ssh
 
 import (
+	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
@@ -12,10 +14,16 @@ import (
 	"os/user"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/juju/errors"
+	"github.com/juju/mutex"
 	"github.com/juju/utils"
+	"github.com/juju/utils/clock"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/knownhosts"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 const sshDefaultPort = 22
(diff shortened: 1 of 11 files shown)


nskaggs:
!!build!!

nskaggs:
!!build!!

nskaggs:
Ahh right, no buildbot here. This should be mergable now.

axw:
> This is great but I'd like confirmation about why all this is necessary as with Juju we typically already have the exact key we need for a given target. Is mimicking all the OpenSSH behaviour required because of the manual provider?

Yes.

ssh/ssh_gocrypto.go
```diff
@@ -234,10 +256,250 @@ func (c *goCryptoCommand) StderrPipe() (io.ReadCloser, io.Writer, error) {
...
 	userHost := strings.SplitN(s, "@", 2)
 	if len(userHost) == 2 {
 		return userHost[0], userHost[1]
 	}
 	return "", userHost[0]
 }
+
+var (
+	goCryptoKnownHostsMutex sync.Mutex
+	goCryptoKnownHostsFile  string
```
wallyworld: Do these really have to be global? 
goCryptoKnownHostsFile could be an attribute of the goCryptoCommand and passed in along with all the other options in the Options struct

axw: Maybe if we did things over, but we have several globals already, including the client itself (i.e. ssh.DefaultClient). It's not worth the effort to try and change it around right now.

axw:
$$merge$$

axw:
$$merge$$

axw:
$$merge$$

axw:
Build failed: Tests failed

PR #272, opened 2017-04-20
Opensuse support

marcmolla:
## Description of change

This change add the support of OpenSUSE Leap (42 series) to Juju. 

Main reason for the change is that most of our NVF run on top of OpenSUSE/SLES and I was doing some hands-on for evaluating Juju as a generic VNF-M. This PR  (and other for juju/juju) is the result of my work.

With these PR, Juju users are able to deploy OpenSUSE charms in LXD and in manual provisioned OpenSUSE host.

## QA steps

A part of the testing that I added, it is possible to deploy an OpenSUSE charm:
https://github.com/marcmolla/juju-OpenSUSE/tree/master/charms/test-opensuseleap

using a modified OpenSUSE image:
https://github.com/marcmolla/juju-OpenSUSE/blob/master/lxd-opensuse-42.2-image.md

I also describe how I tested the software locally:
https://github.com/marcmolla/juju-OpenSUSE/blob/master/test-compiled-Juju.md

For all those testing we also need another PR to juju/juju that I am going to create -> https://github.com/juju/juju/pull/7257

## Documentation changes

If this PR is accepted, we should include reference to OpenSUSE in general doc.

## Bug reference

N/A

PR #277, opened 2017-04-24
Support OpenSUSE series

axw:
This change add the support of OpenSUSE Leap (42 series)
to the utils packages; os, series and packaging.

Supersedes #272.


axw:
$$merge$$

PR #279, opened 2017-05-05
call Sync() before Close() 

gabriel-samfira:
On windows, disk cache and buffers are enabled by default. In the event of a power failure any file that has not yet flushed its contents to disk will become corupt. This leads to state files and agent configs being filled with garbage data.

fixes bug: https://bugs.launchpad.net/juju/+bug/1691193

### DIFF
--- a/file.go
+++ b/file.go
@@ -125,10 +125,15 @@ func AtomicWriteFileAndChange(filename string, contents []byte, change func(*os.
 	if _, err := f.Write(contents); err != nil {
 		return fmt.Errorf("cannot write %q contents: %v", filename, err)
 	}
+	if err := f.Sync(); err != nil {
+		return err
+	}
+	if err := f.Close(); err != nil {
+		return err
+	}
 	if err := change(f); err != nil {
 		return err
 	}
-	f.Close()
 	if err := ReplaceFile(f.Name(), filename); err != nil {
 		return fmt.Errorf("cannot replace %q with %q: %v", f.Name(), filename, err)
 	}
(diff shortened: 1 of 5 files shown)


tar/tar.go
```diff
@@ -155,6 +155,12 @@ func createAndFill(filePath string, mode int64, content io.Reader) error {
 	if err != nil {
 		return fmt.Errorf("cannot set proper mode on file %q: %v", filePath, err)
 	}
+	if err := fh.Sync(); err != nil {
```
babbageclunk: I started to say that it would be better to put this and the close into the defer and make them avoid clobbering any other errors being returned while still returning any errors from them, but actually it's unwieldy enough that I don't think it's worth doing.

file.go
```diff
@@ -125,10 +125,15 @@ func AtomicWriteFileAndChange(filename string, contents []byte, change func(*os.
 	if _, err := f.Write(contents); err != nil {
 		return fmt.Errorf("cannot write %q contents: %v", filename, err)
 	}
+	if err := f.Sync(); err != nil {
+		return err
+	}
+	if err := f.Close(); err != nil {
+		return err
+	}
```
babbageclunk: In that case though the interface should really be changed - the change func should be passed the file name instead of a closed `*File`. There's only one use of `AtomicWriteFileAndChange` in the whole codebase, so it's probably ok to update it.

gabriel-samfira:
$$merge$$

file.go
babbageclunk: Sorry, I missed this - your change here looks great though. Thanks!

PR #280, opened 2017-05-26
parallel: return error when no tries made

rogpeppe:
If for some reason no tries have been made,
rather than returning (nil, nil), we return (nil, ErrNothingStarted)
so we preserve the invariant that if there's no error,
there's a valid result value.

### DIFF
--- a/parallel/try.go
+++ b/parallel/try.go
@@ -12,8 +12,9 @@ import (
 )
 
 var (
-	ErrStopped = errors.New("try was stopped")
-	ErrClosed  = errors.New("try was closed")
+	ErrStopped        = errors.New("try was stopped")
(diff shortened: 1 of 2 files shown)


parallel/try.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 var (
-	ErrStopped = errors.New("try was stopped")
-	ErrClosed  = errors.New("try was closed")
+	ErrStopped        = errors.New("try was stopped")
+	ErrClosed         = errors.New("try was closed")
+	ErrNothingStarted = errors.New("no try attempts were started")
```
jameinel: This reads a little weird.
no attempts were started
?

rogpeppe:
I guess this ain't gonna happen.

PR #282, opened 2017-06-15
add ContextWithTimeout and ContextWithDeadline

rogpeppe:
The stdlib context package doesn't make it easy to use
deadlines and timeouts with mock clocks. These
new functions mimic context.WithTimeout and context.WithDeadline
but using mockable clocks instead.

This does come at some cost - an extra goroutine will
be used when none was needed previously.


rogpeppe:
$$merge$$

rogpeppe:
$$merge$$

PR #283, opened 2017-06-16
return early when deadline already passed in ContextWithDeadline

rogpeppe:
This doesn't make any semantic difference, but is
slightly more efficient.


rogpeppe:
$$merge$$

rogpeppe:
$$merge$$

PR #284, opened 2017-07-29
Keep 1.6 compatibility with context

albertodonato:
Use the x/net/context library instead of the builtin one, to keep compatibilty with go1.6

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,10 +1,11 @@
 package utils
 
 import (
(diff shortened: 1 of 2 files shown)


albertodonato:
@rogpeppe thanks, should be fixed now

rogpeppe:
$$merge$$

PR #287, opened 2017-10-07
Add GetIPv6Address function

wupeka:
Add functions to get IPv6 address of an interface and to get IPv4 (preffered) or IPv6 (fallback) of an interface, to be used by juju LXD provider.

### DIFF
--- a/network.go
+++ b/network.go
@@ -29,6 +29,23 @@ func GetIPv4Address(addresses []net.Addr) (string, error) {
 	return "", fmt.Errorf("no addresses match")
 }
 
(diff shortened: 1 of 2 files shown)


network.go
```diff
@@ -44,3 +61,23 @@ func GetAddressForInterface(interfaceName string) (string, error) {
 	}
 	return GetIPv4Address(addrs)
 }
+
+// GetV4OrV6AddressForInterface looks for the network interface
+// and returns preferrably the IPv4 address, and if it doesn't
```
babbageclunk: Preferably

wupeka:
$$merge$$

PR #288, opened 2017-10-25
series: don't consider pre-release series LTS

axw:
If a series says LTS, but is pre-release, don't
record it as LTS. This is a bit hackish, but
stops us from selecting it as a default series.

### DIFF
--- a/series/series_linux.go
+++ b/series/series_linux.go
@@ -4,13 +4,13 @@
 package series
 
 import (
-	"bufio"
-	"errors"
+	"encoding/csv"
 	"fmt"
-	"io"
 	"os"
 	"strings"
+	"time"
 
+	"github.com/juju/errors"
 	jujuos "github.com/juju/utils/os"
 )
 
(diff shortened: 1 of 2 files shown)


jameinel:
$$merge$$

jameinel:
looking here I only see 2.1.2 for precise
http://streams.canonical.com/juju/tools/streams/v1/com.ubuntu.juju-released-tools.json

dhenrich:
Let's make sure we do this in a way that we never have to fix this again. This seems to have been a reoccurring theme over the years.

hloeung:
Would it be better to work out the default distro to use based off the streams image cache instead (so cache of either internal streams for private clouds such as our ProdStack or public streams.c.c)?

On new release, distro-info-data may have been updated on said host but it's possible that said "cloud" may not have the image sync'ed and available.

axw:
@hloeung yes, I think so: if the user hasn't specified a series, then use the most recent LTS available in both the image and agent repositories.

PR #290, opened 2017-11-22
debugstatus: use gopkg.in/httprequest.v1

rogpeppe:
This makes the debugstatus package use the latest stable version of the httprequest package.


rogpeppe:
$$merge$$

PR #291, opened 2017-12-20
debugstatus: add context to checker

rogpeppe:
This means that we can pass the context through from
the handler to the checker without needing to create
a new debugstatus handler for every request.


rogpeppe:
$$merge$$

PR #293, opened 2018-01-25
Migrate to upstream mgo.

ExternalReality:


rogpeppe:
This doesn't go nearly far enough, I'm afraid.
There are a bunch of other mgo references in utils (mgokv too, for example), some repos that utils depends on also use mgo and should be changed first, and the dependencies.tsv file needs to be updated (godeps ./... > dependencies.tsv).

I recommend using govers (github.com/rogpeppe/govers) to update dependencies - it will also warn you if a dependency is using the package you're wanting to change too.

    govers -m gopkg.in/mgo.v2 github.com/juju/mgo

wallyworld:
superceded by #314

PR #294, opened 2018-02-02
ctxutil: new package

rogpeppe:
We implement a Join function for contexts, which can be
useful when contexts from two origins arrive in the same
place (for example when a long-lived context passed
to a worker interacts with a context created for a single
HTTP request).


rogpeppe:
Useful, but doesn't really belong in juju/utils.

PR #295, opened 2018-02-05
Add ExpandPath w/ tests. Returns abs path for supplied path.

Veebers:
Provide a method that returns an absolute path from the given path.

This will be used (at least initially) to resolve https://bugs.launchpad.net/juju/+bug/1745031 by removing the requirement that the provided path be an absolute path.

### DIFF
--- a/file.go
+++ b/file.go
@@ -11,6 +11,8 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+
+	"github.com/juju/errors"
 )
 
 // UserHomeDir returns the home directory for the specified user, or the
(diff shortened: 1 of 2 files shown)


Veebers:
[GCE Cred bug: 1745031](https://trello.com/c/f1Gf4xgg/178-gce-cred-bug-1745031)

file_test.go
```diff
@@ -73,6 +73,66 @@ func (*fileSuite) TestNormalizePath(c *gc.C) {
...
+		path: "~foobar/path",
+		err:  ".*" + utils.NoSuchUserErrRegexp,
+	}} {
+		c.Logf("test %d: %s", i, test.path)
+		actual, err := utils.ExpandPath(test.path)
+		if test.err != "" {
+			c.Check(err, gc.ErrorMatches, test.err)
+		} else {
+			c.Check(err, gc.IsNil)
+			c.Check(actual, gc.Equals, test.expected)
```
howbazaar: Perhaps also add...
```go
c.Check(filepath.IsAbs(actual) , jc.IsTrue)
```

Veebers:
$$merge$$

file.go
```diff
@@ -51,6 +53,23 @@ func NormalizePath(dir string) (string, error) {
...
 }
 
+// ExpandPath normalises a path and expands it to an absolute form (if it's
+// not already an absolute path).
+func ExpandPath(path string) (string, error) {
+	normPath, err := NormalizePath(path)
+	if err != nil {
+		return "", errors.Annotate(err, "unable to normalise file path")
+	}
+	if !filepath.IsAbs(normPath) {
```
rogpeppe: Why not just use filepath.Abs(normPath) directly? It's more likely to work correctly on platforms other than Unix.

Also, the docs should probably mention Normalize by name - it's not necessarily clear that "normalising" a path implies expanding the `~` prefix.

PR #298, opened 2018-03-20
keyvalues.Join

cmars:
Adds a utility function that joins string maps into a slice of "key=value" string representations for the purpose of sorting by and displaying string map fields.

This function is not intended to be a perfect complement to Parse.

### DIFF
--- a/keyvalues/keyvalues.go
+++ b/keyvalues/keyvalues.go
@@ -8,6 +8,7 @@ package keyvalues
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


nskaggs:
You should mention this is enablement for metrics labels.

keyvalues/keyvalues.go
```diff
@@ -39,3 +40,14 @@ func Parse(src []string, allowEmptyValues bool) (map[string]string, error) {
 	}
 	return results, nil
 }
+
+// Join returns a sorted slice containing each item in the given map, as
+// "key=value" strings.
+func Join(m map[string]string) []string {
```
nskaggs: Is there a reason to sort?

keyvalues/keyvalues.go
```diff
@@ -39,3 +40,14 @@ func Parse(src []string, allowEmptyValues bool) (map[string]string, error) {
...
+	var result []string
+	for k, v := range m {
+		result = append(result, fmt.Sprintf("%s=%s", k, v))
```
nskaggs: Do we want to enforce uniqueness?

keyvalues/keyvalues.go
cmars: Repeatable output, since map ordering is not guaranteed.

keyvalues/keyvalues.go
cmars: Not sure what you mean.. map keys are unique.

cmars:
sometimes a little code duplication isn't such a bad thing, when the intents are different...

PR #302, opened 2018-08-08
adding support for osx mojave

urosj:
To test, you need osx 10.14 mojave.


urosj:
$$merge$$

howbazaar:
We actually moved this out into the juju/os package, and these are left-over because we couldn't rip them out yet. Can you update the juju/os version too plz?

PR #303, opened 2018-08-20
Use juju clock

howbazaar:
Remove the clock interface and type from utils.

These are now available in "github.com/juju/clock".

Drive by fix for some licences.


dependencies.tsv
```diff
@@ -1,12 +1,13 @@
 github.com/juju/ansiterm	git	b99631de12cf04a906c1d4e4ec54fb86eae5863d	2016-09-07T23:45:32Z
-github.com/juju/cmd	git	ad2437ef0ef282ae26e482eb1e44c02532bae1ba	2017-06-22T12:53:07Z
+github.com/juju/clock	git	d293bb356ca441f16b59562295cca74478966757	2018-05-24T02:22:03Z
```
babbageclunk: Why so many dependency changes? Is that because `juju/clock` uses more recent versions of these packages?

howbazaar: I was just bringing the project up to what we have with Juju. The only real one that needed to change was the retry package, but I didn't see harm in updating them all.

howbazaar:
$$merge$$

PR #305, opened 2019-12-09
support multiple feature flags

nammn:
support multiple feature flags by using variadic functions to ensure backward compatibility.

Tests should show patch change.

### DIFF
--- a/featureflag/flags.go
+++ b/featureflag/flags.go
@@ -60,17 +60,22 @@ func setFlags(val string) {
 	}
 }
 
-// Enabled is used to determine if a particular feature flag is enabled for
+// Enabled is used to determine if a particular feature flag/s is/are enabled for
 // the process.
-func Enabled(flag string) bool {
+func Enabled(givenFlags ...string) bool {
 	flaglock.Lock()
 	defer flaglock.Unlock()
-	flag = strings.TrimSpace(strings.ToLower(flag))
-	if flag == "" {
-		// The empty feature is always enabled.
-		return true
+	for _, value := range givenFlags {
+		value = strings.TrimSpace(strings.ToLower(value))
+		if value == "" {
+			// The empty feature is always enabled.
+			return true
+		}
+		if flags.Contains(value) {
+			return true
+		}
 	}
-	return flags.Contains(flag)
+	return false
 }
 
 // All returns all the current feature flags.
--- a/featureflag/flags_test.go
+++ b/featureflag/flags_test.go
@@ -45,3 +45,14 @@ func (s *flagSuite) TestEnabled(c *gc.C) {
 	c.Assert(featureflag.Enabled("Magic"), jc.IsTrue)
 	c.Assert(featureflag.Enabled(" MAGIC "), jc.IsTrue)
 }
+
+// Assume we migrated from an old version to a new one.
+// Out of UX reasons we try to support multiple featureflags for the same feature as below.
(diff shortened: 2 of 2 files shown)


nammn:
Locally they pass. Grumpig does not have the utilities to run those tests.
```
....
bin/sh: 1: zip: not found
...
exec: "/usr/local/bin/mongod": stat /usr/local/bin/mongod: no such file or directory
```

featureflag/flags.go
```diff
@@ -60,17 +60,22 @@ func setFlags(val string) {
 	}
 }
 
-// Enabled is used to determine if a particular feature flag is enabled for
+// Enabled is used to determine if a particular feature flag/s is/are enabled for
```
SimonRichardson: Can we talk about flags in plural, considering we're changing this.

featureflag/flags.go
```diff
@@ -60,17 +60,22 @@ func setFlags(val string) {
...
-func Enabled(flag string) bool {
+func Enabled(givenFlags ...string) bool {
 	flaglock.Lock()
 	defer flaglock.Unlock()
-	flag = strings.TrimSpace(strings.ToLower(flag))
-	if flag == "" {
-		// The empty feature is always enabled.
-		return true
+	for _, value := range givenFlags {
+		value = strings.TrimSpace(strings.ToLower(value))
```
SimonRichardson: Nit: Can we call this `trimmedValue` that way we don't overwrite a shadowed variable.

featureflag/flags_test.go
```diff
@@ -45,3 +45,14 @@ func (s *flagSuite) TestEnabled(c *gc.C) {
 	c.Assert(featureflag.Enabled("Magic"), jc.IsTrue)
 	c.Assert(featureflag.Enabled(" MAGIC "), jc.IsTrue)
 }
+
+// Assume we migrated from an old version to a new one.
+// Out of UX reasons we try to support multiple featureflags for the same feature as below.
+func (s *flagSuite) TestEnabledMultiple(c *gc.C) {
```
SimonRichardson: Can we add a negative test this one.

```go
c.Assert(featureflag.Enabled("fred", "blogs"), jc.IsFalse)
```


```go
c.Assert(featureflag.Enabled("something", "magic"), jc.IsTrue)
```

nammn:
@SimonRichardson @manadart I don't really have a preference as both approaches are fine to me. We can loop @manadart in and see whats his opinion is on either:

## 1.)
Supporting multiple flags as an or bool statement:
```
Enabled(givenFlags ...string) 
```
#### Adv.
- easy to implement
- easy to use in develop and backport

#### Disadv.
- can be confusing because having multiple flags for "one" flag
- may lead to unexpected behaviour if not understood
## 2.) 
Add an alias kind statement
```
s.PatchEnvironment("JUJU_TESTING_FEATURE", "generations")
featureflag.SetFlagAlias("generations", "branches")

featureflag.Enabled("branches") // true
```

#### Adv.
- makes it more clear that one is "only" a alias because of upgrading reasons
#### Disadv.
- more implementation complexity for juju/utils as I suspect from first glance that we need to change our datastructure from a `string set` something like a [array string map/struct]
- little more code (trivial) on the side of juju/juju as we need to have a place which get's called before the cli get's called.

nammn:
As of discussion with @mitechie @SimonRichardson @manadart im closing this in favor of the more simplistic change in `juju/juju` as alternative feature flags are only a one time thing.

PR #306, opened 2020-01-14
Enhance GetHTTPClient to accept optional ca certificates.

hmlanigan:
Add GetHTTPClientWithCerts as an http client constructor allowing ca certs.

Due to dependency issues with github.com/juju/juju, change names.v2 dependency to names.v3.


http.go
```diff
@@ -5,17 +5,14 @@ package utils
...
+	"crypto/x509"
 	"encoding/base64"
 	"fmt"
 	"net"
 	"net/http"
 	"strings"
-	"sync"
 )
 
-var insecureClient = (*http.Client)(nil)
```
hmlanigan: Unused.

http.go
```diff
@@ -43,13 +40,38 @@ const (
...
 		return GetValidatingHTTPClient()
 	}
 	return GetNonValidatingHTTPClient()
 }
 
+// GetHTTPClientWithCerts returns a new http.Client that verifies the
+// server's certificate chain and hostname depending on arguments and
+// adds ca certificates to the client. Returns nil if no certificates
+// provided.
+func GetHTTPClientWithCerts(verify SSLHostnameVerification, certs []string) *http.Client {
```
achilleasa: Given that the `GetHTTPClient` signature is essentially the same, would it be worth making this private so we can force package importers to go through `GetHTTPClient`?

hmlanigan:
Looks like the gitHub-check-merge-juju-utils could use some love.  All of the items in the job which fail, not including the license things, pass for me locally.  And the 3 that fail for me locally, pass in the job: httpDialSuite.TestDefaultClientAllowAccess, httpDialSuite.TestInsecureClientAllowAccess, httpDialSuite.TestSecureClientAllowAccess

http.go
hmlanigan: done

http.go
```diff
@@ -42,19 +42,19 @@ const (
 // non validating client depending on the value of verify.
 func GetHTTPClient(verify SSLHostnameVerification, certs ...string) *http.Client {
 	if len(certs) > 0 {
-		return GetHTTPClientWithCerts(verify, certs)
+		return getHTTPClientWithCerts(verify, certs)
 	}
 	if verify == VerifySSLHostnames {
 		return GetValidatingHTTPClient()
```
achilleasa: Since you are making changes here, I have the same question about `GetValidatingHTTPClient` and `GetNonValidatingHTTPClient`. Are we invoking these functions outside of this package? If not, we can probably also make them internal and have `GetHTTPClient` as the entry point.

hmlanigan: They are both directly used by juju/juju

hmlanigan:
$$merge$$

hmlanigan:
!!build!!

hmlanigan:
rebased to master to pull in Makefile changes etc.

hmlanigan:
$$merge$$

hmlanigan:
$$merge$$

PR #308, opened 2020-02-04
update imports to utils.v2 and remote set pkg

hmlanigan:


howbazaar:
If we are clearing out old stuff, we should be able to remove deque/deque_test.go and the series directory as that should be juju/os/series now.

hmlanigan:
Versioning utils wasn't strictly necessary for what I needed to happen.  It was solved a different way already.

It'd be nice to do, however not trivial to get all the related dependency's sorted.    Ran into a panic trying to bring into juju:

 go run -v ./generate/schemagen/schemagen.go ./apiserver/facades/schema.json
panic: protocol file already registered

goroutine 1 [running]:
net/http.(*Transport).RegisterProtocol(0x448b260, 0x26018bd, 0x4, 0x29c4440, 0xc0001c8310)
	/snap/go/5245/src/net/http/transport.go:577 +0x342
github.com/juju/juju/vendor/github.com/juju/utils.registerFileProtocol(...)
	/home/heather/work/src/github.com/juju/juju/vendor/github.com/juju/utils/http.go:27
github.com/juju/juju/vendor/github.com/juju/utils.init.0()
	/home/heather/work/src/github.com/juju/juju/vendor/github.com/juju/utils/http.go:22 +0xc0
exit status 2

PR #310, opened 2020-04-13
Fix WinRM Run() function

gabriel-samfira:
This change makes the ```Run()``` function rely on the exit code of the command being run, instead of the ```err``` variable returned by the winrm module.

In the past, some ```powershell``` versions did not properly set the exit code on all versions of windows. As a result, we would get an error, but the exit code was being set to ```0```. Back then, the only way to tell if an error occurred, was to look at ```stderr```, which is bad.

Since then, updates have fixed those kinds of scenarios, and we need to look at the actual exit code.

In the case of running powershell over WinRM, some powershell commandlets will generate a progress bar, which unfortunately will output the progress to ```stderr```. This means that successful commands will report errors, which makes the manual provider unusable with Windows.

This is the first patch in fixing the already existing manual provider support for Windows machines.

### DIFF
--- a/winrm/winrm.go
+++ b/winrm/winrm.go
@@ -211,8 +211,11 @@ func NewClient(config ClientConfig) (*Client, error) {
 // If the Run successfully executes it returns nil
 func (c *Client) Run(command string, stdout io.Writer, stderr io.Writer) error {
 	logger.Debugf("Runing cmd on WinRM connection %q", command)
-	_, err := c.conn.Run(command, stdout, stderr)
-	if err != nil {
+	exitCode, err := c.conn.Run(command, stdout, stderr)
+	if exitCode != 0 {
+		if err == nil {
+			err = errors.Errorf("exit status %d", exitCode)
+		}
 		return errors.Annotatef(err, "cannot run WinRM command %q", command)
 	}
 	return nil


winrm/winrm.go
```diff
@@ -211,9 +211,13 @@ func NewClient(config ClientConfig) (*Client, error) {
 // If the Run successfully executes it returns nil
 func (c *Client) Run(command string, stdout io.Writer, stderr io.Writer) error {
 	logger.Debugf("Runing cmd on WinRM connection %q", command)
-	_, err := c.conn.Run(command, stdout, stderr)
-	if err != nil {
-		return errors.Annotatef(err, "cannot run WinRM command %q", command)
+	exitCode, err := c.conn.Run(command, stdout, stderr)
+	if exitCode != 0 {
+		var cmdErr error = err
```
howbazaar, line 216: Why create `cmdErr` here? Doesn't seem like you really need it.

winrm/winrm.go
```diff
@@ -211,9 +211,13 @@ func NewClient(config ClientConfig) (*Client, error) {
...
+		if cmdErr == nil {
+			cmdErr = errors.New(fmt.Sprintf("exit status %d", exitCode))
```
howbazaar, line 218: Should use errors.Errorf.

```go
if exitCode != 0 {
    if err == nil {
        err = errors.Errorf("exit status %d", exitCode)
    }
    return errors.Annotatef(err, "cannot run WinRM command %q", command)
}

gabriel-samfira:
Done. Thanks.

winrm/winrm.go
gabriel-samfira: Isolation is getting to me.

gabriel-samfira:
> The changes look good. Are there any tests for this now? Is the `conn` instance or the `Run` method able to be mocked out for testing?

I can attempt to refactor the ```Client``` struct to allow us to swap the WinRM ```conn``` instance, with a mocked one, if you think it's worth it.

gabriel-samfira:
$$merge$$

PR #311, opened 2020-04-23
Add option to set feature flags from more than one env var

wallyworld:
Enhance SetFlagsFromEnvironment() so that you can pass in more than one env var from which to extract feature flags.
Same with SetFlagsFromRegistry().


wallyworld:
$$merge$$

PR #314, opened 2021-03-05
Update some sbpackages to use github.com/juju/mgo/v2 instead of gopkg.in/mgo.v2

wallyworld:
Update some sbpackages to use github.com/juju/mgo/v2 instead of gopkg.in/mgo.v2


wallyworld:
$$merge$$

PR #317, opened 2022-01-30
Home() returns real home, not snap home

wallyworld:
When running inside a strictly confined snap, `$HOME` is set to a path under `~/snap/foo` which is accessible to the snap without needing to use the `home` interface. However, looking up the home directory of a named user using the Go library APIs still returns the real home. And things like `NormalisePath()` will also use the real home if `~user/path` is used, but not `~/path`.

So for consistency, `Home()` is tweaked to always return the real home. If we want to use the confined home, we can add extra logic for that if/when needed. NB A strictly confined juju will use the real home.

### DIFF
--- a/home_unix.go
+++ b/home_unix.go
@@ -1,5 +1,6 @@
 // Copyright 2013 Canonical Ltd.
 // Licensed under the LGPLv3, see LICENCE file for details.
+//go:build !windows
 // +build !windows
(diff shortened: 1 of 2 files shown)


benhoyt (approved):

wallyworld:
$$merge$$

PR #319, opened 2022-02-02
Fix setting a fake home for tests

wallyworld:
The previous patch to handle SNAP_REAL_HOME didn't account for set but empty values. The result was lxd tests in juju failing.
Here we use the relevant env var if it exists, even if empty.


SimonRichardson (approved):

wallyworld:
$$merge$$

wallyworld:
$$merge$$

PR #322, opened 2022-09-08
[JUJU-1750] Add riscv among supported architectures

cderici:
Adds the `RISC-V` to the list of architectures that Juju recognize. Also added the `risc`, `risc-v64`, `risc-V64` variants.

Note that this does **not** add the `RISC-V32`. I'm not sure if we need it.

### DIFF
--- a/arch/arch_test.go
+++ b/arch/arch_test.go
@@ -40,6 +40,9 @@ func (s *archSuite) TestNormaliseArch(c *gc.C) {
 		{"ppc64le", "ppc64el"},
 		{"ppc64", "ppc64el"},
 		{"s390x", "s390x"},
+		{"risc", "riscv64"},
+		{"risc-v64", "riscv64"},
+		{"risc-V64", "riscv64"},
 	} {
 		arch := arch.NormaliseArch(test.raw)
 		c.Check(arch, gc.Equals, test.arch)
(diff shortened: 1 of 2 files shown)


arch/arch_test.go
```diff
@@ -40,6 +40,9 @@ func (s *archSuite) TestNormaliseArch(c *gc.C) {
 		{"ppc64le", "ppc64el"},
 		{"ppc64", "ppc64el"},
 		{"s390x", "s390x"},
+		{"risc", "riscv64"},
```
hmlanigan, line 43: per https://wiki.debian.org/Multiarch/Tuples#riscv64-linux-gnu, there is a ```riskv64``` we should ensure to capture as well

cderici: Really? with a `k` ? I don't see it anywhere

hmlanigan: `riscv64` - autocorrect me

hmlanigan: the comment really applies to the MustCompile.

cderici:
Note that the failing test is unrelated to this change and also fails on the master branch. We need to fix it separately.

cderici:
@jameinel do you have any additional comments/thoughts about this? (not the failing CI test, but adding the `riscv64`)

cderici:
> Note that the failing test is unrelated to this change and also fails on the master branch. We need to fix it separately.

The fix is up for the tests #323.

manadart (approved):
Looks like it should do the trick.

PR #323, opened 2022-09-13
Fix macthed error messages in unit tests

cderici:
Looks like a bunch of auto generated error messages have changed in the past and the unit tests weren't updated accordingly, so the CI tests are failing. This change fixes it by updating the matched error messages in the unit tests.


cderici:
Looks like some of the tests are only showing in the CI and not in my local run, annoying.

cderici:
[cryptossh.NewServerConn(netconn, s.cfg)](https://github.com/juju/utils/blob/c3fbc78a33b005ee2f033760e19b97d172208873/ssh/ssh_gocrypto_test.go#L46) intermittently returns an `EOF`.

PR #324, opened 2023-01-20
[WIP] Generate Ed25519 ssh keys instead of RSA-2048

jack-w-shaw:
These keys are more secure, smaller and result in quicker speed

Unfortunately, (unlike other keys) the Go standard library doesn't properly marshal Ed25519 keys for SSH. Do this ourselves

### QA Steps

- Delete your `~/.local/share/juju/ssh/juju_id_rsa*` keys
- Compile juju with this branch in place of juju/utils
- Bootstrap a controller and verify Ed25519 keys are generated, and `~/.local/share/juju/ssh/juju_id_rsa.pub` grants access via SSH

### DIFF
--- a/ssh/generate.go
+++ b/ssh/generate.go
@@ -4,48 +4,16 @@
 package ssh
 
 import (
+	"crypto/ed25519"
 	"crypto/rand"
-	"crypto/rsa"
-	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	mathrand "math/rand"
 	"strings"
 
 	"github.com/juju/errors"
 	"golang.org/x/crypto/ssh"
 )
-
-// rsaGenerateKey allows for tests to patch out rsa key generation
-var rsaGenerateKey = rsa.GenerateKey
-
-// KeyBits is used to determine the number of bits to use for the RSA keys
-// created using the GenerateKey function.
-var KeyBits = 2048
-
-// GenerateKey makes a 2048 bit RSA no-passphrase SSH capable key.  The bit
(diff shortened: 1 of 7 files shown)


hpidcock:
From what I gather ed25519 shouldn't be our default since AFAIK its not FIPS approved. We should either default to ECDSA P384 or 3072bit RSA.

ssh/generate.go
```diff
@@ -4,62 +4,123 @@
...
-	if err != nil {
-		return "", "", errors.Trace(err)
+// MarshalEd25519PrivateKey formats an Ed25519 private key into a OpenSSH
+// compatible format
+//
+// NOTE(jack-w-shaw, 2023-01-20) Ideally we would a standard library function
+// to do this, as we can with RSA and ECDSA algorithms. But this doesn't seem
+// possible for Ed25519 as an exception. My implementation is based on
+// https://github.com/mikesmitty/edkey
+func MarshalEd25519PrivateKey(privateKey ed25519.PrivateKey, publicKey ed25519.PublicKey, comment string) []byte {
```
hpidcock, line 66: OpenSSH supports PKCS8 since like v8.1. We don't need to write OpenSSH keys. See what I wrote in the `pki/ssh` package for generating ssh host_keys.

ssh/generate.go
```diff
@@ -4,62 +4,123 @@
 package ssh
```
hpidcock, line 4: this generation work done in `generate.go` should really come from the new `pki/ssh` package. I'd prefer to retire this whole `ssh` package.

PR #325, opened 2023-01-23
[JUJU-2465] Increase SSH rsa keys size to 3072 bits

jack-w-shaw:
2048 is no longer considered completely safe, so increase the size to the new industry standard

3072 is a good balance between high security and performance (without changing the algorithm)

### QA Steps

Compile into juju and verify juju generates 3072-bit RSA keys. You can do this by deleting (backing up if necessary) `$HOME/.local/share/juju/ssh/juju_id_rsa*` and bootstrapping a new controller


anvial (approved):
QA passed, LG2M

PR #335, opened 2024-02-09
Fix generation of ssh id25519 key pairs

wallyworld:
The generation of the id25519 ssh keys was incorrect, such that they could not be used to ssh.

This PR fixes that and also ensure the generated filenames are correct.


hpidcock (approved):
Annoying that openssh has taken so long to support pkcs#8 ed25519 keys, this is only a problem prior to openssh 9.6

PR #336, opened 2024-03-06
Add check for new line is ssh key

Aflynn50:
The function `ParseAuthorisedKey` does not check for new lines in its input. The description above says the function checks "a line from an authorized key file" and the definition for an authorized key in `man sshd` includes the text:
> The optional comment field continues to the end of the line.
Meaning that the comment should not wrap over lines.

A check is added for a newline in the input.

This fixes:
- https://github.com/juju/terraform-provider-juju/issues/363

## QA
Replaced package with local branch in juju 3.3. In go.mod put:
- `replace github.com/juju/utils/v3 => ../utils`
- `go mod tidy`
Ran unit tests for places that use `ParseAuthorizedKey`:
```
go test ./worker/authenticationworker/...
go test ./apiserver/facades/client/keymanager
```

## Other improvements

This PR also fixes some flaky tests by stopping them from using the hosts terminal, and adds a `.gitignore`.

### DIFF
--- a/ssh/authorisedkeys.go
+++ b/ssh/authorisedkeys.go
@@ -55,6 +55,9 @@ func authKeysDir(username string) (string, error) {
 // authorized_keys file and returns the constituent parts.
 // Based on description in "man sshd".
 func ParseAuthorisedKey(line string) (*AuthorisedKey, error) {
+	if strings.Contains(line, "\n") {
+		return nil, errors.NotValidf("newline in authorized_key %q", line)
+	}
 	key, comment, _, _, err := ssh.ParseAuthorizedKey([]byte(line))
 	if err != nil {
 		return nil, errors.Errorf("invalid authorized_key %q", line)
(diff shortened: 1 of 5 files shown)


ssh/authorisedkeys.go
```diff
@@ -55,6 +55,9 @@ func authKeysDir(username string) (string, error) {
 // authorized_keys file and returns the constituent parts.
 // Based on description in "man sshd".
 func ParseAuthorisedKey(line string) (*AuthorisedKey, error) {
+	if strings.Contains(line, "\n") {
+		return nil, errors.Errorf("newline in authorized_key %q", line)
```
hmlanigan, line 59: I'd make this an errors.NotValid error.

Aflynn50: Changed

hmlanigan (approved):
LGTM

PR #339, opened 2025-03-25
feat: allow specifying the file to add the keys to and deleting all keys by force

ale8k:
To enable the sshsession worker work for the SSH proxy, we need to write authorized_keys to a separate file other than the default one that juju uses. This allows us to keep Juju working as is and also enable the new way of SSHing, ensuring that they do not conflict with each other. We also need to allow the deletion of ALL kets within a file, and this is guarded by a force flag that will only be used from the sshsession worker.

### DIFF
--- a/ssh/authorisedkeys.go
+++ b/ssh/authorisedkeys.go
@@ -197,33 +197,7 @@ func AddKeys(user string, newKeys ...string) error {
 	if err != nil {
 		return err
 	}
-	for _, newKey := range newKeys {
-		fingerprint, comment, err := KeyFingerprint(newKey)
-		if err != nil {
-			return err
-		}
-		if comment == "" {
-			return errors.Errorf("cannot add ssh key without comment")
-		}
-		for _, key := range existingKeys {
-			existingFingerprint, existingComment, err := KeyFingerprint(key)
-			if err != nil {
-				// Only log a warning if the unrecognised key line is not a comment.
-				if key[0] != '#' {
-					logger.Warningf("invalid existing ssh key %q: %v", key, err)
-				}
-				continue
-			}
-			if existingFingerprint == fingerprint {
-				return errors.Errorf("cannot add duplicate ssh key: %v", fingerprint)
-			}
-			if existingComment == comment {
-				return errors.Errorf("cannot add ssh key with duplicate comment: %v", comment)
-			}
-		}
-	}
-	sshKeys := append(existingKeys, newKeys...)
-	return writeAuthorisedKeys(user, authKeysFile, sshKeys)
+	return addKeys(user, authKeysFile, newKeys, existingKeys)
 }
 
 // DeleteKeys removes the specified ssh keys from the authorized ssh keys file for user.
@@ -232,48 +206,11 @@ func AddKeys(user string, newKeys ...string) error {
 func DeleteKeys(user string, keyIds ...string) error {
 	keysMutex.Lock()
 	defer keysMutex.Unlock()
-	existingKeyData, err := readAuthorisedKeys(user, authKeysFile)
+	existingKeys, err := readAuthorisedKeys(user, authKeysFile)
 	if err != nil {
 		return err
 	}
-	// Build up a map of keys indexed by fingerprint, and fingerprints indexed by comment
-	// so we can easily get the key represented by each keyId, which may be either a fingerprint
-	// or comment.
-	var keysToWrite []string
-	var sshKeys = make(map[string]string)
-	var keyComments = make(map[string]string)
-	for _, key := range existingKeyData {
-		fingerprint, comment, err := KeyFingerprint(key)
-		if err != nil {
-			logger.Debugf("keeping unrecognised existing ssh key %q: %v", key, err)
-			keysToWrite = append(keysToWrite, key)
-			continue
-		}
-		sshKeys[fingerprint] = key
-		if comment != "" {
-			keyComments[comment] = fingerprint
-		}
-	}
-	for _, keyId := range keyIds {
-		// assume keyId may be a fingerprint
-		fingerprint := keyId
-		_, ok := sshKeys[keyId]
-		if !ok {
-			// keyId is a comment
-			fingerprint, ok = keyComments[keyId]
-		}
-		if !ok {
-			return errors.Errorf("cannot delete non existent key: %v", keyId)
-		}
-		delete(sshKeys, fingerprint)
-	}
-	for _, key := range sshKeys {
-		keysToWrite = append(keysToWrite, key)
-	}
-	if len(keysToWrite) == 0 {
-		return errors.Errorf("cannot delete all keys")
-	}
-	return writeAuthorisedKeys(user, authKeysFile, keysToWrite)
+	return deleteKeys(user, authKeysFile, existingKeys, keyIds, false)
 }
 
 // ReplaceKeys writes the specified ssh keys to the authorized_keys file for user,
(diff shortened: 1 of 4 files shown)


ssh/authorisedkeys.go
```diff
@@ -31,7 +30,7 @@ var (
 )
 
 const (
-	authKeysFile = "authorized_keys"
+// authKeysFile = "authorized_keys"
```
kian99, line 33: Leftover comment

kian99 (commented):
Since this change isn't backwards compatible, have you considered creating functions like `AddKeysWithFile` or something similar that `AddKeys` calls using the default filename, then you can call `AddKeysWithFile` in the new places you need to.

ssh/authorisedkeys_test.go
SimoneDutto, line 1: please add also a test where you write to two different files, in this way we are "sure" that your implementation with the file specified actually goes to two different files

SimoneDutto (commented):
I agree with Kian, if you create another API in this pkg you don't have to worry about migrating juju codebase and potentially other codebases that are relying on this feature.

ssh/authorisedkeys.go
```diff
@@ -15,10 +15,9 @@ import (
 	"sync"
 
 	"github.com/juju/errors"
-	"github.com/juju/loggo/v2"
+	"github.com/juju/loggo"
+	"github.com/juju/utils/v3"
```
wallyworld, line 19: The utils import needs to stay where it was.

But this is the wrong branch, it needs to be v3 branch if we want to use these changes in juju 3.6
The master branch is for v4

wallyworld (changes requested):
As others have said, we can't break compatibility.
Since the changes will be purely additive, we can do them on v3 and release as a new minor version v3.3.0
The PR is targetted against master when it needs to be v3

ssh/authorisedkeys_test.go
```diff
@@ -287,3 +290,39 @@ func (s *AuthorisedKeysKeysSuite) TestConcatAuthorisedKeys(c *gc.C) {
...
+	key2 := sshtesting.ValidKeyTwo.Key + " user@host"
+	err = ssh.AddKeysToFile(testSSHUser, alternativeKeysFile3, []string{key2})
+	c.Assert(err, jc.ErrorIsNil)
+
+	list2, err := ssh.ListKeysFromFile(testSSHUser, alternativeKeysFile3, ssh.FullKeys)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(list2, gc.DeepEquals, []string{key2})
+}
+
+func (s *AuthorisedKeysKeysSuite) TestDeleteKeysFromFile(c *gc.C) {
```
wallyworld, line 312: We want 2 tests - one for deleting some keys, another for deleting all keys

ssh/authorisedkeys_test.go
```diff
@@ -287,3 +290,39 @@ func (s *AuthorisedKeysKeysSuite) TestConcatAuthorisedKeys(c *gc.C) {
...
+	list1, err := ssh.ListKeysFromFile(testSSHUser, alternativeKeysFile2, ssh.FullKeys)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(list1, gc.DeepEquals, []string{key1})
+
+	err = ssh.DeleteKeysFromFile(testSSHUser, alternativeKeysFile2, []string{sshtesting.ValidKeyOne.Fingerprint})
+	c.Assert(err, jc.ErrorIsNil)
+
+	emptyList, err := ssh.ListKeysFromFile(testSSHUser, alternativeKeysFile2, ssh.FullKeys)
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(emptyList, gc.DeepEquals, []string(nil))
```
wallyworld, line 326: `c.Assert(emptyList, gc.HasLen, 0)`

wallyworld (approved):
Looks good once comments are addressed.

wallyworld:
Once this lands can it also be forward merged into master (v4) please.

ale8k:
> Once this lands can it also be forward merged into master (v4) please.

Will do, just investigating why CI failing.

ale8k:
@wallyworld I've tried to investigate why CI is failing, the logs aren't being very helpful. I suspect it's the suite at root but when I run that locally, it's fine. When run them all locally it's fine too. I thought perhaps its a race condition, but even with -count 1000, it's 1000/1000. Any idea what's happening here?

ssh/authorisedkeys.go
```diff
@@ -351,3 +351,121 @@ func EnsureJujuComment(key string) string {
 	}
 	return key
 }
+
+// AddKeysToFile adds the specified ssh keys to the specified file for user.
+// Returns an error if there is an issue with *any* of the supplied keys.
+func AddKeysToFile(user, file string, newKeys []string) error {
```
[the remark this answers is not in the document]
ale8k: So I've changed it because the signature can be a bit silly in that:
```go
AddKeysToFile("ubuntu", "myfile") // places myfile -> ...newKeys
```

I'll remove the duplication though.

ssh/authorisedkeys.go
```diff
@@ -351,3 +351,121 @@ func EnsureJujuComment(key string) string {
...
+	sshKeys := append(existingKeys, newKeys...)
+	return writeAuthorisedKeys(user, file, sshKeys)
+}
+
+// DeleteKeysFromFile removes the specified ssh keys from the authorized ssh keys file for user.
+// keyIds may be either key comments or fingerprints.
+// Returns an error if there is an issue with *any* of the keys to delete.
+//
+// Unlike DeleteKeys, this version can delete ALL keys from the target file.
+func DeleteKeysFromFile(user, file string, keyIds []string) error {
```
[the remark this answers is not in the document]
ale8k: As stated in other comment I'm going to leave this so that we don't fall into variadic pitfalls.

ssh/authorisedkeys.go
```diff
@@ -351,3 +351,121 @@ func EnsureJujuComment(key string) string {
...
+		delete(sshKeys, fingerprint)
+	}
+	for _, key := range sshKeys {
+		keysToWrite = append(keysToWrite, key)
+	}
+	return writeAuthorisedKeys(user, file, keysToWrite)
+}
+
+// ListKeys returns either the full keys or key comments from the authorized ssh keys file for user.
+func ListKeysFromFile(user, file string, mode ListMode) ([]string, error) {
```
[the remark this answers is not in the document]
ale8k: Removed duplication & renamed to defaultAuthKeysFile.

ssh/authorisedkeys_test.go
```diff
@@ -287,3 +290,39 @@ func (s *AuthorisedKeysKeysSuite) TestConcatAuthorisedKeys(c *gc.C) {
...
+	c.Assert(err, jc.ErrorIsNil)
+	c.Assert(list2, gc.DeepEquals, []string{key2})
+}
+
+func (s *AuthorisedKeysKeysSuite) TestDeleteKeysFromFile(c *gc.C) {
+	key1 := sshtesting.ValidKeyOne.Key + " user@host"
+	err := ssh.AddKeysToFile(testSSHUser, alternativeKeysFile2, []string{key1})
+	c.Assert(err, jc.ErrorIsNil)
+
+	list1, err := ssh.ListKeysFromFile(testSSHUser, alternativeKeysFile2, ssh.FullKeys)
```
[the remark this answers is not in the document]
ale8k: Agreed, but I liked seeing the keys listed first. I guess it doesn't provide much benefit as ListKeysFromFile and AddKeysToFile might not work correctly, so instead we should write the keys manually. I'll update to use the write method already in this file.

ssh/authorisedkeys_test.go
```diff
@@ -287,3 +290,39 @@ func (s *AuthorisedKeysKeysSuite) TestConcatAuthorisedKeys(c *gc.C) {
 		c.Check(ssh.ConcatAuthorisedKeys(test.a, test.b), gc.Equals, test.result)
 	}
 }
+
+func (s *AuthorisedKeysKeysSuite) TestAddKeysToFile(c *gc.C) {
+	key1 := sshtesting.ValidKeyOne.Key + " user@host"
+	err := ssh.AddKeysToFile(testSSHUser, alternativeKeysFile2, []string{key1})
```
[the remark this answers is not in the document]
ale8k: Bar the signature, done

ssh/authorisedkeys_test.go
ale8k: Done

ale8k:
@wallyworld Harry found the issue, was a * issue from a while back. It's fixed in this PR now. 

I've done all your comments bar the signature and tried to explain my reasoning. Any further issues let me know and I'll change it to be variadic, but I do think it could lead to issues for any future consumers who aren't careful.

